PR #221, opened 2019-10-09
Add listURLs GraphQL endpoint

obukhov:
Added initial endpoint for https://github.com/byliuyang/short/issues/200 
Planning to add tests, but feel free to start review @byliuyang


byliuyang:
I have few ideas. Please let me know what are your thoughts:

1. Create `viewer(authToken: String): User` endpoint ( https://developer.github.com/v4/explorer )
2. Add `URLOwner` attribute to `User`
3. Create `URLOwner`
```graphql
type URLOwner {
    urls: [URL!]
}
```
4. Decouple business logic away from database and delegate filter logic to `usecase`
   - url repository get all urls ( Database query doesn't seem to be efficient. However, this decouple provide us flexibilities adding new feature )
   - urlRetriver filter URLs by user email

byliuyang:
Please add:
-  integration tests for GraphQL endpoint ( should be HTTP test, sorry for not having this due to me tech debts )
- unit tests for URL retriever ( Only for the feature you are working on )
- unit tests for URL repository

byliuyang:
Thank you so much for working on this cool & challenging feature!

byliuyang:
Any updates?

byliuyang:
Hi @obukhov, are you still working on this PR?

obukhov:
Hello! Sorry, I had some busy days (or week) I would be glad to continue working on this.

> Create viewer(authToken: String): User

I like the idea! It also seem to be a common practice when designing the APIs

> type URLOwner 

I am not sure about this one though, are you sure additional layer is needed here. Maybe it rather should be an interface (or just User type should directly have "ownedUrls" field? Can you please explain your idea in more detail?


> Decouple business logic away from database and delegate filter logic to usecase
> url repository get all urls ( Database query doesn't seem to be efficient. However, this decouple provide us flexibilities adding new feature )

Do you mean to load the whole table in memory and then filter by userId there? That would be super inefficient for thousands and for millions of users would be deadly. Don't you think filtering logic is better to delegate database engine (keeping it abstracted, so potentially the engine can be changed)?

I don't think that filtering urls by owner is business logic. At least, not the one that cannot be delegated to the storage engine.

Once again sorry for the delay, and let me know what you think about my comments, so I can continue the implementation.

byliuyang:
Hi @obukhov, I highly agree with all your opinions! I did have the same concerns for Database gateways. For the `URLOwner` one, I saw it at GitHub API v4. Let’s implement all your suggestions! 

I am lucky to have this very technical discussion with you!

byliuyang:
I’ll close this PR for now due to no activity. Please reopen the PR when you want to keep working on it.

PR #245, opened 2019-10-13
Configures Drone CI to enable integration tests

### LINKED ISSUE #204: Configure Drone CI to run DB integration tests
Currently, there are only unit tests running on the CI. There is no integration tests covering the GraphQL and RESTful. To avoid breaking critical features in the future, we want to add integration tests to automatically verify those feature are not breaking when merging to master branch.

## Success criteria
Integration tests are running for the branches that have pull requests opened for.

matthewhartstonge:
### Description
- Currently Drone will only run tests on master on pull request
- Backend requires postgres to store stateful information when performing API requests when integration tests have been created.

### Description
- Drone should run on any branch, on pull request.
- Drone now brings up a postgreSQL service which can be accessed at `postgres`

Fixes #204


matthewhartstonge:
@byliuyang will do :)
To let you know: my next few nights are booked out, but will have something to you within a week. 👌

byliuyang:
@matthewhartstonge Excited to see your design & implementation!

matthewhartstonge:
Cool, so I had a play and came up with the following as a first run:

- Since the go build seemed to all rely on a common base, I decided to make a "go-builder" container that everything can be built from. This means everything has the required dependencies rather than having to pull them each time. But it does mean having to set up an automated docker build for the build container (that way you don't have to worry about maintaining it yourself)
- I noticed the Go service itself, while being on Alpine, actually ran on top of the Golang image, so while I was there I stripped that out to reduce the container size down to ~22MBs down from ~599MB
- From drone I already see I need to correct my branch/pr condition logic :P but it gives you an idea of where I'm heading

Thoughts so far?

byliuyang:
https://github.com/byliuyang/short/pull/292

byliuyang:
I’ll close this PR for now due to no activity. Please reopen the PR when you want to keep working on it.

Issue #281, opened 2019-10-25
Configure Kubernetes


byliuyang:
Short is now depending on Key Gen service. It will also be depending on auth service, feature toggle service, etc... Kubernetes is here to help manage all those services.

byliuyang:
Install & config [microk8s](https://github.com/ubuntu/microk8s) on the server

byliuyang:
Currently, we are using [Microk8s](https://microk8s.io/docs/) for single machine cluster to save money. When we get more traffics in the future, we will switch to full size Kubernetes.

byliuyang:
```yaml
# nginx.yaml
kind: Service
apiVersion: v1
metadata:
  name: ingress
  namespace: ingress
  labels:
    app.kubernetes.io/name: ingress
    app.kubernetes.io/part-of: ingress
spec:
  externalTrafficPolicy: Local
  type: LoadBalancer
  selector:
    app.kubernetes.io/name: ingress
    app.kubernetes.io/part-of: ingress
  ports:
    - name: http
      port: 80
      targetPort: http
    - name: https
      port: 443
      targetPort: https
  externalIPs:
    - load-balancer-ip
```

All domain names should be resolved to `load-balancer-ip`

byliuyang:
<img width="1677" alt="Screen Shot 2019-11-02 at 9 27 43 AM" src="https://user-images.githubusercontent.com/3537801/68074005-66c7e500-fd53-11e9-8f38-1dda58f0a27b.png">

byliuyang:
<img width="1644" alt="Screen Shot 2019-11-11 at 1 35 10 AM" src="https://user-images.githubusercontent.com/3537801/68576763-8631d280-0423-11ea-9405-f0317beb3753.png">

byliuyang:
# Staging
<img width="676" alt="Screen Shot 2019-11-11 at 1 36 01 AM" src="https://user-images.githubusercontent.com/3537801/68576828-b5484400-0423-11ea-92e2-385a0fe81349.png">

# Production
<img width="709" alt="Screen Shot 2019-11-11 at 1 36 14 AM" src="https://user-images.githubusercontent.com/3537801/68576836-b8433480-0423-11ea-90ed-e8632fca42c4.png">

byliuyang:
#315

PR #292, opened 2019-10-27
Introduce builder pattern for backend and frontend containers

### LINKED ISSUE #290: Multi-stage build for frontend and backend
I found out that one can use either image from the hub or pre-build frontend application locally to pack it to docker. 

That means that you have to have yarn installed locally if you make changes to the frontend.

There is a similar problem with backend. Distribution can carry only binary (which will make it much more lightweight) instead of carrying the whole golang sdk.

I would suggest enabling frontend build with docker multi-stage build for frontend (and possibly for backend). I'm working on PR for that.

obukhov:
Introduces builder pattern for backend and frontend containers (fixes #290)

Here are the benefits:
- For frontend container: it makes it possible to build with no yarn or build directory existing locally.
- For backend container: it optimizes container size (599MB -> 21.8MB)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,12 +2,24 @@ version: '3'
 services:
   frontend:
     image: ${DOCKERHUB_USERNAME}/short-frontend:latest
+    build:
+      context: ./frontend
+      dockerfile: Dockerfile
+      args:
+        REACT_APP_GIT_SHA: ${REACT_APP_GIT_SHA}
+        REACT_APP_GRAPHQL_API_BASE_URL: ${GRAPHQL_API_BASE_URL}
+        REACT_APP_HTTP_API_BASE_URL: ${HTTP_API_BASE_URL}
+        REACT_APP_RECAPTCHA_SITE_KEY: ${RECAPTCHA_SITE_KEY}
     ports:
       - "${WEB_PORT}:80"
     depends_on:
       - backend
+
   backend:
     image: ${DOCKERHUB_USERNAME}/short-backend:latest
+    build:
+      context: ./backend
+      dockerfile: Dockerfile
     ports:
       - "${GRAPHQL_API_PORT}:8080"
       - "${HTTP_API_PORT}:80"
--- a/.env.dist
+++ b/.env.dist
@@ -1,0 +1,22 @@
+WEB_PORT=3000
+HTTP_API_PORT=8082
+GRAPHQL_API_PORT=8083
+
+WEB_FRONTEND_URL=http://localhost:3000
+GRAPHQL_API_BASE_URL=http://localhost:8083
+HTTP_API_BASE_URL=http://localhost:8082
+
+DB_USER=user
+DB_PASSWORD=pass
+DB_NAME=short
+
+REACT_APP_GIT_SHA=dev
+DOCKERHUB_USERNAME=local
+
+RECAPTCHA_SITE_KEY=your_recaptcha_site_key
+RECAPTCHA_SECRET=your_recaptcha_secret
+
+GITHUB_CLIENT_ID=your_Github_client_id
+GITHUB_CLIENT_SECRET= your_Github_client_secret
+
+JWT_SECRET= your_JWT_secret
--- a/README.md
+++ b/README.md
@@ -47,6 +47,11 @@ git clone https://github.com/byliuyang/short.git
    HTTP_API_PORT=80
    GRAPHQL_API_PORT=8080
    ```
+
+Or just copy `.env.dist` to a new name:
+```
+cp .env.dist .env
+```
    
 1. Update `DB_USER`, `DB_PASSWORD`, `DB_NAME`, and `JWT_SECRET` with your own
    configurations.
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.13.1-alpine
+FROM golang:1.13.1-alpine AS builder
 
 WORKDIR /app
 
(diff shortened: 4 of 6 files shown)


byliuyang:
Hi @obukhov , thank you so much for working this optimization! The PR itself looks great. My only concern is that we are reusing the `Dockerfile` and `docker-compose.yml` for continuous deliver and one-click deployment through [CI](https://github.com/byliuyang/short/blob/master/.drone.yml). I wonder will the changes break our CI builds?

byliuyang:
I am now [looking into Kubernetes](https://github.com/byliuyang/short/issues/281) and see whether we can ease deployment.

obukhov:
> My only concern is that we are reusing the Dockerfile and docker-compose.yml for continuous deliver and one-click deployment through CI. I wonder will the changes break our CI builds?

I don't see any breaking changes so far, but I think  have to check more thoroughly to be sure. 
Regarding reusing the same dockerfiles - I think it is totally fine as it actually makes it more "fail-fasty", all the optimisations can be achieved by using multistage and I would even prefer to get rid of Dockerfile-test files and replace it with certain stage in one robust Dockerfile. 

Regarding reusing docker-compose - the current one is more suitable for running in prod as it uses optimised container images (although I agree that it is better to deploy with k8s - can help with that if my help is needed there) - In future I would rather introduce a dev version of the containers (wich uses features of dev builds of react and (if applicable) golang) with the separate docker-compose-dev.yml file.

byliuyang:
Hi @obukhov , I have few questions. We building using https://github.com/byliuyang/short/pull/292/files#diff-4e5e90c6228fd48698d074241c2ba760R5-R12.

The CI is building the production containers here: https://github.com/byliuyang/short/blob/master/.drone.yml#L75-L75

and publishing the generated containers here:
https://github.com/byliuyang/short/blob/master/.drone.yml#L106-L110

The CI then uploads the docker-compose file here: https://github.com/byliuyang/short/blob/master/.drone.yml#L120-L139

At the end, it launches the containers: https://github.com/byliuyang/short/blob/master/.drone.yml#L202-L220

I didn't see the `docker-compose.yml` is referenced during build stages. And the source code is not copied on to production machine. I wonder how are we building the images and launching the containers in this situation.

Please feel free to decouple when you see fit. Yes, I need your help with Kubernetes!

obukhov:
Thanks for the overview of .drone file, I haven't look at it before. Now I think I know how it works.

> I didn't see the docker-compose.yml is referenced during build stages in the CI config file. 

Yeah, for now, docker-compose file is not used for building - it wouldn't be possible before the change I've made (added build key), so it was used only for starting the services (as you mentioned on lines 202 - 220).

> I wonder how are we building the images and how are we launching the containers in this situation.

Currently, the build process is happening (as you said) on line 75 with a direct docker build command. It still will be working after my changes, but there is an alternative now: Instead of providing these arguments inline `-t short-frontend:latest -f frontend/Dockerfile frontend` it is possible to run the build with parameters specified under `build` key in docker-compose.yml file. Then build command would be as simple as `docker-compose build` or `docker-compose build <serviceName>` (where serviceName is either backend or frontend). The only requirement is that runtime container specified for example on line 73 (now it is "docker") should have docker-compose in it (which I am unfortunately not sure about because I don't have much experience with drone-ci).

obukhov:
According to this build: https://ci.time4hacks.com/byliuyang/short/509/1/5 

This:
> The only requirement is that runtime container specified for example on line 73 (now it is "docker") should have docker-compose

is not the case, I'll keep experimenting with that PR and then close it.

obukhov:
Hey wasn't able to make it work yesterday. Please don't merge - I think it is important to change the way it builds not only locally but on ci as well

byliuyang:
@obukhov I have been testing Kubernetes this whole night. Will continue trying it out tomorrow night.

byliuyang:
#281

byliuyang:
The multi stage build looks great! We just to to be careful about docker-compose file. Btw, k8s is on the way to be ready: https://github.com/byliuyang/short/tree/master/k8s

byliuyang:
I am looking forwards to merge your PR whenever it's ready

docker-compose.yml
```diff
@@ -2,12 +2,24 @@ version: '3'
 services:
```
byliuyang: This PR is ready to be merged if we exclude changes to this file

obukhov:
@byliuyang Hey this thing slipped away from me again, sorry. I've update the branch, not sure if there's anything left to fix.

I was trying to enable build through docker-compose, but didn't manage, maybe still worth merging this, although it has some duplication between docker-compose and drone configuration (not good for DRY). I still going to fix that.

byliuyang:
Hi @obukhov , thank you so much for optimizing this. You are definitely one of the most talented contributors to Short among those I have been working with. 

We are on the way migrating to Kubernetes, I would appreciate if you can add production Kubernetes configurations instead. https://github.com/byliuyang/short/tree/master/k8s

Please also join the Slack chat room mentioned in the README so that you can receive deployment status notifications. ( We will together figure out how to allocated per developer testing environment if you are interested in )

byliuyang:
The early this PR is merged, the more network bandwidth and developer time we will save.

byliuyang:
#315 #281

byliuyang:
Merged.

byliuyang:
For your information, this PR reduced the size of the frontend image to [9.18 MB](https://hub.docker.com/layers/byliuyang/short-frontend-staging/latest/images/sha256-8f23865529d168f4b7846378fb2309984c587e8740ee8cf716c68a718eb0cd88) and backend image to [11.61 MB](https://hub.docker.com/layers/byliuyang/short-backend-staging/latest/images/sha256-22d1188edc39ae79e50231afba165eed97a67bb3c084f7f26c7b0d89c4c41f4a).

Issue #299, opened 2019-11-06
Add database attribute indicating whether a short link is public


Coteh:
Depends on #235. This attribute can be added as a new column to the `url` table. By default, short links will be private, so this will be set to false by default. This column should also be non-nullable, as we only expect two states (either it's public or it isn't).

byliuyang:
In my opinion, `private` or `public` is not naturally a property of `url`. How about creating a new table called `public_url`?

Coteh:
#307 has been merged.

Issue #235, opened 2019-10-12
[Feature] Add private / public option when creating short link


byliuyang:
Currently, all the short links created are private to their authors. We want to encourage people to share some of their short links with others to give back to the community.

To achieve this:

- Add a toggle button on the short link creation panel  to allow use to specify whether the short link created will be publicly available. By default, the short link should be private by default.
- Add a database attribute to indicate whether the short link is public.
- Update GraphQL short link creation API with new argument

Please create a separate issue for each litem listed here and link them back to this one.

Coteh:
I can give this one a try.

Coteh:
Issues created.

PR #307, opened 2019-11-07
Add public_url table

Coteh:
Add a schema for public URLs to the database. This is for #299 and is an incremental change to address #235.


backend/app/adapter/migration/8_create_public_url_table.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Up
+CREATE TABLE public_url (
+    alias CHARACTER VARYING,
```
byliuyang:
```suggestion
    alias CHARACTER VARYING(50)  NOT NULL,
```

Coteh:
@byliuyang That makes sense to have it as `varchar(50)`, as `user_url_relation` also has the same restriction on `alias` column. Thanks for the suggestion! Should the `alias` column in `url` table be updated to have this restriction as well?

byliuyang:
Hi @Coteh , yes. Alias in url table should have the same restriction. Could you please open an issue to propose this change? Thanks!

Coteh:
@byliuyang Issue created. #316. Also going to close #299.

Issue #192, opened 2019-10-01
Repeatedly request sign in 


byliuyang:
The sign in dialog shows up repeatedly when trying to create a short link. This happens when user's Github email is empty.

<img width="762" alt="Screen Shot 2019-09-30 at 10 33 19 PM" src="https://user-images.githubusercontent.com/3537801/65936706-59c17a00-e3d2-11e9-8c63-7a81fa333bbc.png">

lawaloy:
I will like to work on this issue

byliuyang:
@lawaloy Thank you!

byliuyang:
@lawaloy Please join this Slack channel to discuss bugs, dev environment setup, tooling, and coding best practices.

byliuyang:
@lawaloy May I have your email?

lawaloy:
Lasei_1992@yahoo.com

Kind Regards,

Oluwaseyi Lawal  
 
  On Fri, Oct 11, 2019 at 12:05 PM, Harry Liu<notifications@github.com> wrote:   
@lawaloy May I have your email?

lawaloy:
> @lawaloy May I have your email?

I was unable to see the link because your message is coming to my email box directly.

byliuyang:
I invited you to Slack and you joined already.

lawaloy:
Okay, thanks. I'm still working on that task.  
 
  On Sat, Oct 12, 2019 at 12:46 PM, Harry Liu<notifications@github.com> wrote:   I invited you to Slack and you joined already.

byliuyang:
Hi @lawaloy, this bug involves redesigning the database tables and changing Github OAuth & Authentication logic. It's very complex. Do you want to try a [different issue](https://github.com/byliuyang/short/issues/246) instead? I am planing to take on this.

lawaloy:
Okay, That is fine.

lawaloy:
Okay, that is fine. I will take on another issue

byliuyang:
@lawaloy Thank you!

byliuyang:
Migrate from email to user ID.

PR #360, opened 2019-11-18
Add SingleSignOn use case

byliuyang:
### Description
Short is using oauth as it's default authentication method.

### Description
Generalize oauth to single sign on.


byliuyang:
#192

PR #363, opened 2019-11-19
Fix error collecting err codes

Coteh:
### Description
When collecting error codes in Url service error handler, it assumes that a `GraphQlError` object will always contain an `extensions` property and causes frontend to crash when `extensions` is missing.
image
See https://github.com/graphql/graphql-js/blob/v14.5.8/src/error/GraphQLError.js#L83

### Description
Now just unknown error will appear.
### Screenshots
image

### DIFF
--- a/frontend/src/service/Error.service.ts
+++ b/frontend/src/service/Error.service.ts
@@ -3,7 +3,9 @@ import { IErr } from '../entity/Err';
 export enum ErrUrl {
   AliasAlreadyExist = 'aliasAlreadyExist',
   UserNotHuman = 'requesterNotHuman',
-  Unauthorized = 'invalidAuthToken'
+  Unauthorized = 'invalidAuthToken',
+  NetworkError = 'networkError',
+  Unknown = 'unknownError',
 }
 
 const unknownErr = {
@@ -33,6 +35,15 @@ const userNotHumanErr = {
                 `
 };
 
+const networkErr = {
+  name: 'Network error',
+  description: `
+              Could not get a response from the server. Please ensure you are
+              connected to the Internet and try again. If the problem persists,
+              please email byliuyang11@gmail.com.
+              `
+};
+
 export class ErrorService {
   getErr(errCode: ErrUrl): IErr {
     switch (errCode) {
@@ -40,6 +51,9 @@ export class ErrorService {
         return aliasNotAvailableErr;
       case ErrUrl.UserNotHuman:
         return userNotHumanErr;
+      case ErrUrl.NetworkError:
+        return networkErr;
+      case ErrUrl.Unknown:
       default:
         return unknownErr;
     }
--- a/frontend/src/service/Error.service.ts
+++ b/frontend/src/service/Error.service.ts
@@ -40,7 +40,7 @@ const networkErr = {
   description: `
               Could not get a response from the server. Please ensure you are
               connected to the Internet and try again. If the problem persists,
-              please email byliuyang11@gmail.com.
+              please email byliuyang11@gmail.com with screenshots or logs.
               `
 };
 
--- a/frontend/src/service/Error.service.ts
+++ b/frontend/src/service/Error.service.ts
@@ -38,9 +38,8 @@ const userNotHumanErr = {
 const networkErr = {
   name: 'Network error',
   description: `
-              Could not get a response from the server. Please ensure you are
-              connected to the Internet and try again. If the problem persists,
-              please email byliuyang11@gmail.com with screenshots or logs.
+              Unable to reach the server. Please double check your Internet connection and try again.
+              If this happens consistently, please email byliuyang11@gmail.com screenshots and the necessary steps to reproduce the error.
               `
 };
 
--- a/frontend/src/service/Error.service.ts
+++ b/frontend/src/service/Error.service.ts
@@ -39,7 +39,7 @@ const networkErr = {
   name: 'Network error',
   description: `
               Unable to reach the server. Please double check your Internet connection and try again.
-              If this happens consistently, please email byliuyang11@gmail.com screenshots and the necessary steps to reproduce the error.
+              If this happens consistently, please email byliuyang11@gmail.com with screenshots and the necessary steps to reproduce the error.
               `
 };
 
--- a/frontend/src/service/Error.service.ts
+++ b/frontend/src/service/Error.service.ts
@@ -1,6 +1,6 @@
 import { IErr } from '../entity/Err';
 
-export enum ErrUrl {
+export enum Err {
   AliasAlreadyExist = 'aliasAlreadyExist',
   UserNotHuman = 'requesterNotHuman',
   Unauthorized = 'invalidAuthToken',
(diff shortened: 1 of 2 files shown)


frontend/src/service/Url.service.ts
```diff
@@ -151,8 +151,8 @@ export class UrlService {
         .catch(({ graphQLErrors, networkError, message }) => {
           console.log(graphQLErrors);
```
byliuyang:
```suggestion
```

frontend/src/service/Url.service.ts
```diff
@@ -151,8 +151,8 @@ export class UrlService {
...
           const errCodes = graphQLErrors.map(
-            (graphQLError: GraphQlError) => graphQLError.extensions.code
-          );
+            (graphQLError: GraphQlError) => (graphQLError.extensions) ? graphQLError.extensions.code : null
```
byliuyang:
```suggestion
            graphQLError: GraphQlError => (graphQLError.extensions) ? graphQLError.extensions.code : null
```

Please add the following above line 153 to handle network error case.

```typescript
if (!graphQLErrors) {
  return reject([networkError]);
}
```

Can we add `Unknown` at [here](https://github.com/byliuyang/short/blob/1aa5c474c273362fb0c8a1a78ad324f1c1b3bdef/frontend/src/service/Error.service.ts#L6-L7) and use enum instead of null at line 154? Even though `extensions` is empty, it's still an graphQL error.

Coteh:
@byliuyang Removing the parentheses from `(graphQLError: GraphQlError)` like in your second suggestion produces a compiler error. Therefore, I kept them in. Also, I tested the network error case by turning off the backend server, it seems as though `graphQlErrors` will be empty instead of a falsy value, so I also added a case for empty length.

Ready for review.

frontend/src/service/Error.service.ts
```diff
@@ -3,7 +3,9 @@ import { IErr } from '../entity/Err';
 export enum ErrUrl {
```
byliuyang: export enum Err

frontend/src/service/Url.service.ts
```diff
@@ -149,9 +149,14 @@ export class UrlService {
           resolve(res.data.authMutation.createURL);
         })
         .catch(({ graphQLErrors, networkError, message }) => {
-          console.log(graphQLErrors);
+          if (!graphQLErrors || graphQLErrors.length === 0) {
+            if (networkError) {
```
byliuyang: Let’s try to reduce the number of nested if statements ( cyclomatic complexity): https://www.google.com/amp/s/www.geeksforgeeks.org/cyclomatic-complexity/amp/

Coteh: Ah I see, so what we have here is a cyclomatic complexity of 4 because we branch and then do a branch inside of that branch (nested if). I have pushed out a fix that should now reduce the cyclomatic complexity to 2 if I am understanding this correctly, as there's now two if statements not nested. I put the check for network error on top, as I believe network errors should be prioritized over GraphQL errors.

frontend/src/service/Error.service.ts
```diff
@@ -33,13 +35,25 @@ const userNotHumanErr = {
                 `
 };
 
+const networkErr = {
+  name: 'Network error',
+  description: `
+              Could not get a response from the server. Please ensure you are
```
byliuyang:
```suggestion
              Unable to reach the server. Please double check your Internet connection and try again.
```

frontend/src/service/Error.service.ts
```diff
@@ -33,13 +35,25 @@ const userNotHumanErr = {
...
+              connected to the Internet and try again. If the problem persists,
```
byliuyang:
```suggestion
              If this happens consistently, please email byliuyang11@gmail.com screenshots and the necessary steps to reproduce the error.
```

frontend/src/service/Error.service.ts
```diff
@@ -33,13 +35,25 @@ const userNotHumanErr = {
...
+              please email byliuyang11@gmail.com with screenshots or logs.
```
byliuyang:
```suggestion
```

frontend/src/service/Url.service.ts
```diff
@@ -149,9 +149,14 @@ export class UrlService {
           resolve(res.data.authMutation.createURL);
         })
         .catch(({ graphQLErrors, networkError, message }) => {
-          console.log(graphQLErrors);
+          if (networkError) {
+            return reject([Err.NetworkError]);
```
byliuyang:
```suggestion
            reject([Err.NetworkError]);
            return;
```

frontend/src/service/Url.service.ts
```diff
@@ -149,9 +149,14 @@ export class UrlService {
...
+          }
+          if (!graphQLErrors || graphQLErrors.length === 0) { 
+            return reject([Err.Unknown]);
+          }
           const errCodes = graphQLErrors.map(
-            (graphQLError: GraphQlError) => graphQLError.extensions.code
+            (graphQLError: GraphQlError) => (graphQLError.extensions) ? graphQLError.extensions.code : Err.Unknown
```
byliuyang: Are parentheses needed for `(graphQLError.extensions)`?

frontend/src/service/Url.service.ts
```diff
@@ -149,9 +149,14 @@ export class UrlService {
           resolve(res.data.authMutation.createURL);
         })
         .catch(({ graphQLErrors, networkError, message }) => {
-          console.log(graphQLErrors);
+          if (networkError) {
+            return reject([Err.NetworkError]);
+          }
+          if (!graphQLErrors || graphQLErrors.length === 0) { 
+            return reject([Err.Unknown]);
```
byliuyang:
```suggestion
           reject([Err.Unknown]);
           return;
```

byliuyang:
Hi thank you very much for working on this feature! Please address the PR comments.

byliuyang:
By the way, I wonder what's going on here?
Screen Shot 2019-11-19 at 8 50 34 PM

Coteh:
> By the way, I wonder what's going on here?
> Screen Shot 2019-11-19 at 8 50 34 PM

I'm not sure what's going on here actually. I had my eye on it, but subsequent error messages had the X as expected.

frontend/src/service/Url.service.ts
Coteh: Nope they can be left out :)

Coteh:
@byliuyang I guess for whatever reason the material icon for the close button did not load in that particular run-through. Do you have any ideas on what it might be? We can move into a new issue for it.

Ready for review.

frontend/src/service/Error.service.ts
```diff
@@ -33,13 +35,24 @@ const userNotHumanErr = {
                 `
 };
 
+const networkErr = {
+  name: 'Network error',
+  description: `
+              Unable to reach the server. Please double check your Internet connection and try again.
+              If this happens consistently, please email byliuyang11@gmail.com screenshots and the necessary steps to reproduce the error.
```
byliuyang:
```suggestion
              If this happens consistently, please email byliuyang11@gmail.com with screenshots and the necessary steps to reproduce the error.
```

PR #371, opened 2019-11-23
Replace account interface with account provider

byliuyang:
### Description
Use cases access user account data through Account interface. 

### Description
Use cases access user account data through Account Provider


backend/app/usecase/account/provider_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}{
+		{
+			name:           "successfully created account",
+			users:          []entity.User{},
+			email:          "alpha@example.com",
+			userName:       "Alpha",
+			expectedHasErr: false,
+		},
+		{
+			name: "successfully created account",
```
byliuyang:
```suggestion
			name: "account exists",
```

PR #409, opened 2019-12-04
Add IsUserIDExist to UserFake and fix UpdateUserID bug

byliuyang:
### Description
1) Fake user repository doesn't have a way to check whether an user with given ID exists. 
2) User ID is not properly updated when invoking UpdateUserID method.

### Description
1) Add IsUserIDExist to UserFake
2) Fix UpdateUserID bug

### DIFF
--- a/backend/app/usecase/repository/user_fake.go
+++ b/backend/app/usecase/repository/user_fake.go
@@ -12,7 +12,7 @@ type UserFake struct {
 	users []entity.User
 }
 
-// IsEmailExist checks whether an user with given email exist in the repository.
+// IsEmailExist checks whether an user with given email exists in the repository.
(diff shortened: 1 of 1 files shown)


backend/app/usecase/repository/user_fake.go
```diff
@@ -22,6 +22,16 @@ func (u UserFake) IsEmailExist(email string) (bool, error) {
...
+func (u UserFake) IsUserIDExist(userID string) bool {
+	for _, user := range u.users {
+		if user.ID == userID {
+			return true
+		}
+	}
+	return false
+}
+
 // GetUserByEmail find an user with a given email.
```
byliuyang:
```suggestion
// GetUserByEmail finds an user with a given email.
```

PR #411, opened 2019-12-04
Add account linker

byliuyang:
### Description
External account is mapped to Short's internal account through email. However, some identity provider only reveals user's public email, resulting in sign in failure.

### Description
Add account linker to map external account to Short's internal account through user IDs.


byliuyang:
https://github.com/byliuyang/short/issues/192

byliuyang:
https://github.com/byliuyang/short/issues/192

PR #421, opened 2019-12-12
Add Account for Google OAuth

dawntomm:
### Description
Add account for Google OAuth

In order to support Google sign in, we are mapping user's Google account to user's Short account through user's email. After retrieving access token from Google OAuth IdentityProvider, Account sends access token to Google API to fetch user's email and name.


backend/app/adapter/google/account.go
```diff
@@ -0,0 +1,52 @@
...
+// GetSingleSignOnUser retrieves user's email and name from Google API.
+func (a Account) GetSingleSignOnUser(accessToken string) (entity.SSOUser, error) {
+	type response struct {
+		Email string `json:"email"`
+		Name  string `json:"name"`
+		ID    string `json:"id"`
+	}
+
+	var res response
+
```
byliuyang:
```suggestion
```

byliuyang:
#423

PR #439, opened 2019-12-17
Add comment to error code enums

byliuyang:


byliuyang:
https://github.com/byliuyang/short/issues/203

PR #452, opened 2019-12-24
Improve wording of "aliasNotAvailable" error message

nathan815:
The original message said "generate _a_ available alias" instead of "generate _an_ available alias" but I ended up further improving the wording. Simple change 🙂


byliuyang:
Love this change! Thank you!

PR #458, opened 2019-12-27
Support running integration test on CI

byliuyang:


byliuyang:
https://github.com/byliuyang/short/issues/204

PR #476, opened 2020-01-08
Implement CreateMapping for GithubSSOSql repository

byliuyang:
Add uniqueness constraints to `github_user_id` and `short_user_id`. 
One Github account should be mapped to only one Short account. Vise versa.


byliuyang:
https://github.com/short-d/short/issues/192

PR #478, opened 2020-01-08
Support retrieving Github user ID

byliuyang:


byliuyang:
https://github.com/short-d/short/issues/192

PR #479, opened 2020-01-09
Add gateway integration test for Google account API

dawntomm:

### DIFF
--- a/backend/app/adapter/google/account_integration_test.go
+++ b/backend/app/adapter/google/account_integration_test.go
@@ -1,3 +1,5 @@
+// +build !integration all
+
 package google
 
 import (
@@ -74,12 +76,14 @@ func TestAccount_GetSingleSignOnUser(t *testing.T) {
 					mdtest.Equal(t, "https://openidconnect.googleapis.com/v1/userinfo", req.URL.String())
 					mdtest.Equal(t, "GET", req.Method)
 					mdtest.Equal(t, "application/json", req.Header.Get("Accept"))
+					mdtest.Equal(t, "Bearer access_token", req.Header.Get("Authorization"))
 
 					return testCase.httpResponse, testCase.httpErr
 				})
 			googleAccount := NewAccount(httpRequest)
 
 			gotSSOUser, err := googleAccount.GetSingleSignOnUser("access_token")
+
 			if testCase.expectHasErr {
 				mdtest.NotEqual(t, nil, err)
 				return
--- a/backend/app/adapter/google/account_integration_test.go
+++ b/backend/app/adapter/google/account_integration_test.go
@@ -1,4 +1,4 @@
-// +build !integration all
+// +build integration all
 
 package google
 
--- a/backend/app/adapter/google/account_integration_test.go
+++ b/backend/app/adapter/google/account_integration_test.go
@@ -1,0 +1,91 @@
+package google
+
+import (
+	"bytes"
+	"errors"
+	"io/ioutil"
+	"net/http"
+	"testing"
+
+	"github.com/short-d/app/mdtest"
+	"github.com/short-d/short/app/entity"
+)
+
+func TestAccount_GetSingleSignOnUser(t *testing.T) {
+	testCases := []struct {
+		name            string
+		httpResponse    *http.Response
+		httpErr         error
+		expectHasErr    bool
+		expectedSSOUser entity.SSOUser
+	}{
+		{
+			name:         "invalid access token",
+			httpResponse: nil,
+			httpErr:      errors.New("invalid access token"),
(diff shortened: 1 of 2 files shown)


backend/app/adapter/google/account_integration_test.go
```diff
@@ -0,0 +1,91 @@
+package google
```
alldroll: Please add `// +build integration all` build tag.

All integration tests are run by this [script](https://github.com/short-d/short/blob/master/backend/scripts/integration-test)

backend/app/adapter/google/account_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+		},
+	}
+
+	for _, testCase := range testCases {
+		testCase := testCase
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+			httpRequest := mdtest.NewHTTPRequestFake(
+				func(req *http.Request) (response *http.Response, e error) {
+					mdtest.Equal(t, "https://openidconnect.googleapis.com/v1/userinfo", req.URL.String())
```
alldroll: If you checked request URL here, maybe it will be a good idea to check that `access token` has been added to headers?

backend/app/adapter/google/account_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/short-d/app/mdtest"
+	"github.com/short-d/short/app/entity"
+)
+
+func TestAccount_GetSingleSignOnUser(t *testing.T) {
+	testCases := []struct {
+		name            string
+		httpResponse    *http.Response
+		httpErr         error
+		expectHasErr    bool
```
alldroll: `expectHasErr` equals to `true` only when `httpErr` is not `nil`. Maybe we can replace these 2 variables with one variable `expectedErr`?

Also, you can use [errors.Is](https://golang.org/pkg/errors/#Is)

backend/app/adapter/google/account_integration_test.go
byliuyang: Please add the build tag @alldroll suggested. Build tag is one of the techniques we use to select which tests to run. We are separating unit tests and integration tests because when unit tests fail ( check whether the module itself is implemented correctly ), there is no point of running integration tests ( check whether two modules are interacting correctly). This allows us to find failures in the shortest time.

backend/app/adapter/google/account_integration_test.go
byliuyang: We should check access token in header because it's part of Google's API spec

backend/app/adapter/google/account_integration_test.go
byliuyang: Currently, we don't have a error types defined for account API errors. I wonder whether we can still use`Is` here. I am reading [this](https://blog.golang.org/go1.13-errors) blog.

byliuyang:
Code climate is passing now but not synced with the PR. This is a new package we started having test coverage, so it's okay for the project test coverage to drop temporarily.

backend/app/adapter/google/account_integration_test.go
dawntomm: addressed comment, thanks for letting me know

backend/app/adapter/google/account_integration_test.go
dawntomm: can we add something like assert.hasErr(), assert.NoErr()

backend/app/adapter/google/account_integration_test.go
dawntomm: addressed comment

backend/app/adapter/google/account_integration_test.go
```diff
@@ -0,0 +1,95 @@
+// +build !integration all
```
byliuyang:
```suggestion
// +build integration all
```
You should mark it as an integration test instead of unit test.

backend/app/adapter/google/account_integration_test.go
byliuyang: @dawntomm Unfortunately no. This is prohibited by Google's Go coding best practices.

backend/app/adapter/google/account_integration_test.go
dawntomm: if that's the case, github/account_integration_test.go line 1 should be changed as well, right

byliuyang: @dawntomm Good catch! Will open a PR to address that soon.

byliuyang: https://github.com/short-d/short/pull/481

PR #486, opened 2020-01-20
Add Presenter interface

byliuyang:
### Description
Currently, there are business and presentation logics co-living inside `routing` package.

### Description
Presentation logic is abstracted our through `Presenter` interface, allowing business logics to be reused for sending text messages in the future.


byliuyang:

Issue #495, opened 2020-01-20
[Refactor] Move business logics from routing package to usecase package  


byliuyang:
**What is the problem?**
Currently, there are business and presentation logics co-living inside routing package.

**Your solution**
Abstract out presentation logic through Presenter interface and move business rules into use cases.

**Additional context**
![image](https://user-images.githubusercontent.com/3537801/72750316-67616200-3b71-11ea-87eb-1e5fc1fdd949.png)

byliuyang:
This is making the code unnecessarily complex at the current stage of the project. Closing this issue for now.

PR #503, opened 2020-01-30
Move request Github sign in logic to use case

byliuyang:
### Description
Github sign in redirection logic is invoking HTTP APIs directly inside [routing](https://github.com/short-d/short/blob/master/backend/app/adapter/routing/handle.go#L57-L65) package. This prevents the sign in logic from being reused for non-HTTP views. It also significantly increased the complexity of `routing` package as more routes being added.

### Description
Github sign in redirection logic is moved into `usecase` package. Interactions with views are abstracted out through `Presenter` interface.

### DIFF
--- a/backend/app/usecase/presenter_mock.go
+++ b/backend/app/usecase/presenter_mock.go
@@ -52,6 +52,7 @@ func (m mockPresenter) ShowInvalidAuthTokenError() {
 
 func newMockPresenter() mockPresenter {
 	return mockPresenter{
+		showHomeCallArgs:         make([]showHomeCallArgs, 0),
 		showUserHomeCallArgs:     make([]showUserHomeCallArgs, 0),
 		showExternalPageCallArgs: make([]showExternalPageCallArgs, 0),
 	}
--- a/backend/app/usecase/presenter_mock.go
+++ b/backend/app/usecase/presenter_mock.go
@@ -1,0 +1,58 @@
+package usecase
+
+type showHomeCallArgs struct {
+	authToken string
+}
+
+type showUserHomeCallArgs struct {
+	authToken string
+}
+
+type showExternalPageCallArgs struct {
+	link string
+}
+
+var _ Presenter = (*mockPresenter)(nil)
+
+type mockPresenter struct {
+	showHomeCallArgs         []showHomeCallArgs
+	showUserHomeCallArgs     []showUserHomeCallArgs
+	showExternalPageCallArgs []showExternalPageCallArgs
+}
+
+func (m *mockPresenter) ShowHome() {
+	m.showHomeCallArgs = append(m.showHomeCallArgs, showHomeCallArgs{})
+}
+
+func (m *mockPresenter) ShowUserHome(authToken string) {
+	m.showUserHomeCallArgs = append(
+		m.showUserHomeCallArgs,
+		showUserHomeCallArgs{authToken: authToken},
+	)
+}
+
+func (m mockPresenter) ShowAliasNotFound() {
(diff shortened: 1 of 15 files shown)


backend/app/usecase/presenter_mock.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (m mockPresenter) ShowInvalidAuthTokenError() {
+	panic("implement me")
+}
+
+func newMockPresenter() mockPresenter {
+	return mockPresenter{
+		showUserHomeCallArgs:     make([]showUserHomeCallArgs, 0),
+		showExternalPageCallArgs: make([]showExternalPageCallArgs, 0),
```
alldroll: Here I see 2 problems:
* `showHomeCallArgs` is not initialized
* we should negotiate how to initialize empty slice:
```
1. make([]showUserHomeCallArgs, 0)
2. []showUserHomeCallArgs{}
```

backend/app/usecase/usecase.go
```diff
@@ -0,0 +1,58 @@
...
+) {
+	if s.authenticator.IsSignedIn(authToken) {
+		presenter.ShowHome()
+		return
+	}
+	signInLink := identityProvider.GetAuthorizationURL()
+	presenter.ShowExternalPage(signInLink)
+}
+
+type GithubIDProvider service.IdentityProvider
```
alldroll: What is the purpose of this type?

backend/app/usecase/usecase.go
```diff
@@ -0,0 +1,58 @@
...
+type GithubSSOAccount service.SSOAccount
```
alldroll: This type is unused, isn't it?

backend/app/usecase/usecase_test.go
```diff
@@ -0,0 +1,195 @@
...
+			useCase := newUseCase(
+				testCase.now,
+				testCase.existingURLs,
+				testCase.tokenValidDuration,
+				testCase.githubIDProvider,
+			)
+			presenter := newMockPresenter()
+			useCase.RequestGithubSignIn(testCase.authToken, &presenter)
+
+			mdtest.Equal(t, testCase.expectedShowUserHomeCallArgs, presenter.showUserHomeCallArgs)
```
alldroll: We should check `expectedShowHomeCallArgs` value

backend/app/usecase/presenter_mock.go
byliuyang: @alldroll I saw most of the docs uses `make` to init empty struct. What are your thoughts?

byliuyang: Done.

backend/app/usecase/usecase.go
byliuyang: @alldroll The explicit type casting to help autocomplete and to prevent error. In the following PR, there will be `GoogleIDProvider`, and `FacebookIDProvider`. Trying to distinguish them.

backend/app/usecase/usecase.go
byliuyang: Removed.

backend/app/usecase/usecase_test.go
byliuyang: Done.

backend/app/usecase/presenter_mock.go
alldroll: I guess there is no difference. We just have to decide what to use ^)

byliuyang:
Addressed all code review comments.

byliuyang:

PR #505, opened 2020-02-01
Move request Facebook sign in to use case

byliuyang:
### Description
Github sign in redirection logic is invoking HTTP APIs directly inside [routing](https://github.com/short-d/short/blob/master/backend/app/adapter/routing/handle.go#L57-L65) package. This prevents the sign in logic from being reused for non-HTTP views. It also significantly increased the complexity of `routing` package as more routes being added.

### Description
Facebook sign in redirection logic is moved into `usecase` package. Interactions with views are abstracted out through `Presenter` interface.


byliuyang:

PR #511, opened 2020-03-04
deploy

kapillamba4:


byliuyang:
Let's hold off deploying until [this bug](https://github.com/short-d/short/issues/512) is fixed

PR #518, opened 2020-03-05
Replace KeyGenerator interface with concrete implementation

### LINKED ISSUE #513: [Refactor] Remove KeyGenerator interface
**What is the problem?**
Currently NewKey() under KeyGenerator is abstracted out through interface.

**Your solution**
Replace the unnecessary interface with concrete implementation.

dawntomm:
fixes #513


backend/app/usecase/url/urlcreator_test.go
```diff
@@ -6,7 +6,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/short-d/short/app/usecase/service"
+
 	"github.com/short-d/app/mdtest"
+	kgsentity "github.com/short-d/kgs/app/entity"
```
byliuyang: We should NOT import entities from other services.

backend/app/usecase/service/keyfetcher.go
```diff
@@ -1,8 +1,9 @@
 package service
 
-import "github.com/short-d/kgs/app/entity"
+// Key represents unique identifier
+type Key string
```
dawntomm: added a Key type, so that we don't rely on external package

backend/dep/provider/keygen.go
```diff
@@ -8,11 +8,11 @@ import (
 // KeyGenBufferSize specifies the size of the local cache for fetched keys
 type KeyGenBufferSize int
 
-// NewRemote creates Remote with KeyGenBufferSize to uniquely identify
+// NewKeyGenerator creates remote NewKeyGenerator with KeyGenBufferSize to uniquely identify
```
byliuyang:
```suggestion
// NewKeyGenerator creates KeyGenerator with KeyGenBufferSize to uniquely identify
```

backend/app/usecase/service/keyfetcher.go
byliuyang: Should this be part of entity?

backend/app/usecase/account/linker_test.go
```diff
@@ -5,6 +5,8 @@ package account
 import (
 	"testing"
 
+	"github.com/short-d/short/app/usecase/service"
+
```
byliuyang:
```suggestion
```

byliuyang:
Hi Tom, thank you so much for your dedication towards improving the maintainability of our code base! I left few comments on the PR.

backend/app/usecase/service/keyfetcher.go
dawntomm: according to our offline discussion, this should be put in service layer, which is right here

byliuyang: Discussed offline. Keeping Key in service package for now.

PR #525, opened 2020-03-07
Remove unused code

byliuyang:
### Description
There are unused variables in SearchService. 

### Description
The unused variables are removed. They will be added back when the real implementation is provided.


kapillamba4:
They might be needed when we integrate it with api

byliuyang:
I agree. That's why I approved the PR in the first place. However, the
linter is complaining. So I removed them for now. We can add them back when
we are actually integrating with the backend.

PR #538, opened 2020-03-09
[Draft] Added Search API

dawntomm:
#  New Behavior
### Description
Added RESTful API for search.

User should be able to search the public and the private short links they created. In order to support that, we add RESTful API to search for all urls created by a given user.


byliuyang:
Close this PR because the APIs are redesigned.

PR #539, opened 2020-03-09
Added GetByAliases function in db and repository

dawntomm:
### Description
Added GetByAliases function in db and repository.

User should be able to search the public and the private short links they created. In order to support that, we add RESTful API to search for all urls created by a given user. The first step is to add GetByAliases function to fetch URLs from database.


backend/app/adapter/db/url.go
```diff
@@ -96,6 +96,21 @@ WHERE "%s"=$1;`,
 	return url, nil
 }
 
+// GetByAliases finds URLs for a list of aliases
+func (u URLSql) GetByAliases(aliases []string) ([]entity.URL, error) {
+	urls := make([]entity.URL, 0)
```
alldroll: I think it is a good practice to respect [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)

backend/app/adapter/db/url.go
```diff
@@ -96,6 +96,21 @@ WHERE "%s"=$1;`,
...
+
+	for _, alias := range aliases {
+		url, err := u.GetByAlias(alias)
```
alldroll: I think it is a potential bottleneck. Let's imagine we want to fetch 1000 aliases, so we will perform 1000 sequent requests. I think we should do next:

* We implement the ability to pass the list of aliases and perform an operation by one query. ( SQL IN Operator )
* `GetByAlias` should call `GetByAliases` with providing the list of a single element.

In that case, we will have only one place with SQL logic and `GetByAlias` will be a helper

alldroll:
Thank you for your PR!. Please address my comments :)

backend/app/adapter/db/url.go
dawntomm: will change to that

backend/app/adapter/db/url.go
dawntomm: I agree on you with performing an operation by one query with SQL IN Operator. However, I feel that ```GetByAlias``` should stay the same, since it does not need SQL IN Operator, which has a high time complexity.

dawntomm: I have made changes to ```GetByAlias```, please take a look when you get the chance, thank you!

backend/app/adapter/db/url_integration_test.go
```diff
@@ -242,6 +242,117 @@ func TestURLSql_Create(t *testing.T) {
...
+			expectedURLs: []entity.URL{
+				entity.URL{
+					Alias:       "220uFicCJj",
+					OriginalURL: "http://www.google.com",
+					CreatedAt:   nil,
+					ExpireAt:    nil,
+					UpdatedAt:   nil,
+				},
+			},
+		},
```
byliuyang: I think this test case is unnecessary. Are you trying to cover specific errors?

backend/app/adapter/db/url.go
```diff
@@ -96,6 +96,53 @@ WHERE "%s"=$1;`,
...
 }
 
+// GetByAliases finds URLs for a list of aliases
+func (u URLSql) GetByAliases(aliases []string) ([]entity.URL, error) {
+	var urls []entity.URL
+
+	statement := fmt.Sprintf(`
+SELECT "%s","%s","%s","%s","%s" 
+FROM "%s" 
+WHERE "%s" IN ("%v");`,
```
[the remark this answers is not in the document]
byliuyang: https://www.geeksforgeeks.org/mitigation-sql-injection-attack-using-prepared-statements-parameterized-queries/
https://en.wikipedia.org/wiki/Prepared_statement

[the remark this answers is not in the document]
byliuyang: https://www.manniwood.com/2019_04_28/go_pg_stmt.html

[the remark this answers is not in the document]
dawntomm: I just changed to parameterized query, please take a look

backend/app/adapter/db/url_integration_test.go
dawntomm: trying to test urls with different time field, but you are right, I just removed it

backend/app/adapter/db/.env
```diff
@@ -1,7 +1,7 @@
 DB_HOST=localhost
-DB_PORT=5432
+DB_PORT=5433
 DB_USER=postgres
-DB_PASSWORD=password
-DB_NAME=short-test
+DB_PASSWORD=123456
+DB_NAME=tom
 
 MIGRATION_ROOT=./migration
```
byliuyang: Please revert changes to this file. This is a template used for integration testing.

backend/app/adapter/db/url.go
```diff
@@ -98,28 +99,42 @@ WHERE "%s"=$1;`,
...
+	for i := range aliases {
+		params = append(params, fmt.Sprintf("$%d", i+1))
+	}
+	parameterStr := strings.Join(params, ", ")
+
+	// create a list of interface{} to hold aliases for db.Query()
+	aliasesInterface := []interface{}{}
+	for _, alias := range aliases {
+		aliasesInterface = append(aliasesInterface, alias)
+	}
```
byliuyang: I couldn't find whether did you `tx.Prepare`.

dawntomm: I just added prepare function in the latest commit

backend/app/adapter/db/url.go
```diff
@@ -96,6 +97,73 @@ WHERE "%s"=$1;`,
...
+
+	var urls []entity.URL
+
+	statement := fmt.Sprintf(`
+SELECT "%s","%s","%s","%s","%s" 
+FROM "%s"
+WHERE "%s" IN (%s);`,
```
byliuyang: Let's not do unsanitized string formatting here. It's serious security risk.

backend/app/adapter/db/url.go
```diff
@@ -96,6 +97,73 @@ WHERE "%s"=$1;`,
...
+		parameterStr,
+	)
+
+	stmt, err := u.db.Prepare(statement)
+	if err != nil {
+		return urls, err
+	}
+	defer stmt.Close()
+
+	rows, err := stmt.Query(aliasesInterface...)
```
byliuyang: Couldn't find the `for` loop and `QueryRow` in the PR. Please checkout this example mentioned in one of the previous articles:

```go
db, err := sql.Open("postgres", "postgres://postgres:postgres@localhost/postgres?sslmode=disable")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()
	stmt, err := db.Prepare("select count(*) from pg_stat_activity where datname = $1")
	if err != nil {
		log.Fatal(err)
	}
	defer stmt.Close()
	var num int
	start := time.Now()
	for i := 0; i < 1000; i++ {
		err = stmt.QueryRow("postgres").Scan(&num)
	}
	elapsed := time.Since(start)
	fmt.Printf("got: %d in %s\n", num, elapsed)
```

backend/app/adapter/db/url.go
byliuyang: Discussed offline. I miss read the code.

backend/app/adapter/db/url.go
dawntomm: according to official document for function ```QueryRow```:
_**// If the query selects no rows, the *Row's Scan will return ErrNoRows.
// Otherwise, the *Row's Scan scans the first selected row and discards
// the rest.**_

dawntomm: it seems that the document is only selecting a single row for testing purpose.
Here we want to select a series of rows, I believe we have to use ```Query```

byliuyang: Please read this whole thread: https://github.com/golang/go/issues/5171

alldroll: @byliuyang In that thread, they talk about bulk operations `a way to insert multiple rows at once`.

Our case is a standard SQL query, which requires to call `Query` and pass the list of args.
Please read this http://go-database-sql.org/prepared.html

> Go creates prepared statements for you under the covers. A simple db.Query(sql, param1, param2), for example, works by preparing the sql, then executing it with the parameters and finally closing the statement.

byliuyang: @alldroll I want to make everything explicit to avoid magics.
@dawntomm I don't want to block your changes. Let's add a TODO here to further discuss our SQL handling strategy. I'll approve this PR after the TODO is added.

backend/app/adapter/db/url.go
```diff
@@ -96,6 +97,73 @@ WHERE "%s"=$1;`,
...
 
+// GetByAliases finds URLs for a list of aliases
+func (u URLSql) GetByAliases(aliases []string) ([]entity.URL, error) {
+	// create a string for aliases parameters
+	// it looks like: $1, $2, $3, ......
+	params := make([]string, 0, len(aliases))
+	for i := range aliases {
+		params = append(params, fmt.Sprintf("$%d", i+1))
+	}
+	parameterStr := strings.Join(params, ", ")
```
byliuyang: Let's extract this to a standalone function called `composeParamList(numParams int) string`

backend/app/adapter/db/url.go
dawntomm: @byliuyang Just added a TODO

backend/app/adapter/db/url.go
dawntomm: addressed comment

backend/app/adapter/db/url.go
dawntomm: @byliuyang sorry I still did not get why you think we should use QueryRow instead of Query, if we use QueryRow then it will cost us more connections with DB, we could definitely do performance test, but intuitively I feel it's slower

byliuyang: @dawntomm All queryRows in a single prepare statement share one connection.

backend/app/adapter/db/url.go
```diff
@@ -96,6 +97,79 @@ WHERE "%s"=$1;`,
...
+
+	// create a list of interface{} to hold aliases for db.Query()
+	aliasesInterface := []interface{}{}
+	for _, alias := range aliases {
+		aliasesInterface = append(aliasesInterface, alias)
+	}
+
+	var urls []entity.URL
+
+	// TODO: compare performance between Query and QueryRow, consider switching to QueryRow function in the future
```
byliuyang:
```suggestion
	// TODO: compare performance between Query and QueryRow. Prefer QueryRow for readability
```

backend/app/adapter/db/url.go
```diff
@@ -96,6 +97,79 @@ WHERE "%s"=$1;`,
...
+		url.UpdatedAt = utc(url.UpdatedAt)
+		url.ExpireAt = utc(url.ExpireAt)
+
+		urls = append(urls, url)
+	}
+
+	return urls, nil
+}
+
+// composeParamListString creates a string for aliases parameters, it looks like: $1, $2, $3, ...
```
byliuyang:
```suggestion
// composeParamListString converts an slice to a parameters string with format: $1, $2, $3, ...
```

backend/app/adapter/db/url.go
```diff
@@ -96,6 +97,79 @@ WHERE "%s"=$1;`,
...
+		url.ExpireAt = utc(url.ExpireAt)
+
+		urls = append(urls, url)
+	}
+
+	return urls, nil
+}
+
+// composeParamListString converts an slice to a parameters string with format: $1, $2, $3, ...
+func (u URLSql) composeParamListString(numParams int) string {
```
byliuyang:
```suggestion
func (u URLSql) composeParamList(numParams int) string {
```

backend/app/adapter/db/url.go
```diff
@@ -96,6 +97,79 @@ WHERE "%s"=$1;`,
 	return url, nil
 }
 
+// GetByAliases finds URLs for a list of aliases
+func (u URLSql) GetByAliases(aliases []string) ([]entity.URL, error) {
+	parameterStr := u.composeParamListString(len(aliases))
```
byliuyang:
```suggestion
	parameterStr := u.composeParamList(len(aliases))
```

backend/app/adapter/db/url.go
byliuyang: The type of the return value already clarified it's string.

backend/app/adapter/db/url.go
```diff
@@ -96,6 +97,79 @@ WHERE "%s"=$1;`,
...
+		url.UpdatedAt = utc(url.UpdatedAt)
+		url.ExpireAt = utc(url.ExpireAt)
+
+		urls = append(urls, url)
+	}
+
+	return urls, nil
+}
+
+// composeParamListString converts an slice to a parameters string with format: $1, $2, $3, ...
```
byliuyang:
```suggestion
// composeParamList converts an slice to a parameters string with format: $1, $2, $3, ...
```

byliuyang:
@alldroll Tom addressed all your comments. I'll merge now to unblock him to work on the next component.

Issue #260, opened 2019-10-16
Create RESTful search API in Go


byliuyang:
Please support async and request cancellation based on the examples provided in the comment.
#238

byliuyang:
Here are some references on goroutine:

1. https://talks.golang.org/2012/concurrency.slide#1
2. https://blog.golang.org/context
3. https://blog.golang.org/pipelines

lucifercr07:
@byliuyang will like to take it up

lucifercr07:
Api: /api/search
Header: Auth token
Query params: 
1. Paging: query params `page.size = 10, page.number = 1` (static for now)
2. type: private or public

byliuyang:
@lucifercr07 For now, we can hard code `page.size`. Let open an issue to track `custom page` size feature. Please document the API endpoint with Swagger ( OpenAPI 3 ). Let's also open an new issue to configure Dredd, a language-agnostic command-line tool for validating API description document against backend implementation of the API.

byliuyang:
Let's scope this feature down to easy development progress:

1) Create RESTful API placeholders in `adapter/routing` package
2) Create repository interfaces in `usecase/repo` package. Those interfaces are purely responsible for database access and have nothing to do with search.  Create search functionalities (are dependent on `usecase/repo` interfaces) in `usercase/search` package.
3) Call search use cases in the RESTful API.

Hope this helps.

luisfmelo:
Is this being developed?

lucifercr07:
@luisfmelo yes

byliuyang:
Hi @luisfmelo are there any other issue you are interested?

byliuyang:
For now, please only search private alias

dawntomm:
I will work on it

byliuyang:
https://github.com/short-d/short/tree/search

byliuyang:
Maybe also consider how to build our own Elasticsearch.

byliuyang:
https://www.elastic.co/elasticsearch/

byliuyang:
https://www.educative.io/courses/grokking-the-system-design-interview/xV9mMjj74gE

arberiii:
During my research I found those articles useful:
- The above link mention (Designing Twitter Search)
- https://www.elastic.co/blog/found-elasticsearch-from-the-bottom-up
- https://www.elastic.co/blog/found-sizing-elasticsearch
- https://www.elastic.co/blog/found-elasticsearch-top-down

[...]

PR #551, opened 2020-03-12
Added Search struct to search URLs for a user

dawntomm:
### Description
Added Search struct to search URLs for a user.

User should be able to search the public and the private short links they created. In order to support that, we add RESTful API to search for all urls created by a given user. The second step is to add Search struct, which provides function to search urls for a user.

### DIFF
--- a/backend/app/usecase/search/search.go
+++ b/backend/app/usecase/search/search.go
@@ -1,0 +1,36 @@
+package search
+
+import (
+	"github.com/short-d/short/app/entity"
+	"github.com/short-d/short/app/usecase/repository"
+)
+
+// Search fetches URLs and users from persistent
+// storage, such as database
+type Search struct {
+	userUrlRepo repository.UserURLRelation
+	urlRepo     repository.URL
+}
+
+// SearchForURLs fetches all URLs for a given user
+func (s Search) SearchForURLs(user entity.User) ([]entity.URL, error) {
+	aliases, err := s.userUrlRepo.FindAliasesByUser(user)
+	if err != nil {
+		return nil, err
+	}
+
+	urls, err := s.urlRepo.GetByAliases(aliases)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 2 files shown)


backend/app/usecase/search/search.go
```diff
@@ -0,0 +1,36 @@
...
+// Search fetches URLs and users from persistent
+// storage, such as database
+type Search struct {
+	userURLRepo repository.UserURLRelation
+	urlRepo     repository.URL
+}
+
+// SearchForURLs fetches all URLs for a given user
+func (s Search) SearchForURLs(user entity.User) ([]entity.URL, error) {
+	aliases, err := s.userURLRepo.FindAliasesByUser(user)
```
alldroll: I do concern about the possible high load problems. I don't know we have restrictions for a number of links created by a user, but how to be if a user has 10k links?

I suggest implementing `paginator` in order to make this potential problem less impact.
What are your thoughts about it?

backend/app/usecase/search/search_test.go
```diff
@@ -0,0 +1,71 @@
...
+			name:         "no such user",
+			user:         user_2,
+			expectedURLs: emptyURLs,
+			expectedErr:  nil,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			searchAPI := NewSearch(&urlRepo, &userUrlRepo)
```
alldroll: let's initialize `&urlRepo, &userUrlRepo` inside `Run`.

byliuyang:
#260

byliuyang:
Please support pagination: https://github.com/short-d/short/issues/260#issuecomment-544273964

byliuyang:
Hi @dawntomm, the code coverage is dropping.

byliuyang:
Closing this PR because the APIs will be redesigned to support pagination.

PR #553, opened 2020-03-14
Add Updates Service

kapillamba4:
### Description
Added UpdatesService

### Screenshots
captured


frontend/src/service/Updates.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    return new Promise(async (resolve, reject) => {
+      resolve(await this.invokeLastSeenChangeLogApi());
+    });
+  }
+
+  getChangeLog(): Promise<Array<Update>> {
+    return new Promise(async (resolve, reject) => {
+      resolve(await this.invokeUpdatesApi());
+    });
+  }
```
kapillamba4: I think this violates SRP

frontend/src/dep.ts
```diff
@@ -46,6 +47,7 @@ export function initUIFactory(
   const versionService = new VersionService(envService);
   const store = initStore();
   const searchService = new SearchService();
+  const updatesService = new UpdatesService();
```
rohithbalaji123: I am wondering, instead of injecting the UpdatesService into UIFactory, can we use it as a static class with static methods? I don't see any use case where we will be needing to store user specific states in this service class. What are your thoughts? @kapillamba4 @byliuyang

kapillamba4: @rohithbalaji123  I think it will depend on UserService or AuthService or ErrorService. User need to known in order to fetch its last seen time from backend.

rohithbalaji123: Ohkay, yeah. Makes sense!

byliuyang: @rohithbalaji123 We prefer instance than static so that we can swap implementation during unit testing. We'll soon start adding unit tests.

frontend/src/component/pages/Home.tsx
```diff
@@ -91,6 +93,17 @@ export class Home extends Component<Props, State> {
...
+    const changeLog = await this.props.updatesService.getChangeLog();
+    this.setState({ changeLog });
+    const lastSeenTimestamp = await this.props.updatesService.getLastSeenChangeLog();
+    if (
+      changeLog &&
+      changeLog[0] &&
+      lastSeenTimestamp < changeLog[0].releasedAt
+    ) {
+      this.handleShowChangeLogBtnClick();
+    }
```
byliuyang:
```suggestion
    if (!changeLog) {
        return;
    }
    
    if (changeLog.length < 1) {
        return;
    }
    const lastSeenAt = await this.props.updatesService.getLastSeenAt();
   if (lastSeenAt >= changeLog[0].releasedAt) {
       return;
    )
   this.showChangeLogs();
```

frontend/src/component/pages/Home.tsx
```diff
@@ -32,6 +32,7 @@ import { UrlService } from '../../service/Url.service';
 import { SearchService } from '../../service/Search.service';
 import { Update } from '../../entity/Update';
 import { ChangeLogModal } from '../ui/ChangeLogModal';
+import { UpdatesService } from '../../service/Updates.service';
```
byliuyang:
```suggestion
import { ChangeLogService } from '../../service/Updates.service';
```

frontend/src/component/pages/Home.tsx
byliuyang: Instead of doing this, let's encapsulate the logic inside `ChangeLogService` to reduce the complexity of `Home` page.:

```typescript
if(!this.props.updatesService.hasUpdates()) {
    return false;
}
// setState is async
 this.setState({ changeLog }, ()=>{
      const changeLog = await this.props.updatesService.getChangeLog();
 });
this.showChangeLogs();
```

byliuyang: The we don't need to expose `getLastSeenChangeLog`

PR #560, opened 2020-03-14
Add ChangeLogService

kapillamba4:
### Description
Added ChangeLogService

### Screenshots
captured


frontend/src/component/UIFactory.tsx
```diff
@@ -18,6 +18,7 @@ import { UrlService } from '../service/Url.service';
 import { SearchService } from '../service/Search.service';
 import { SearchBar } from './ui/SearchBar';
 import { ViewChangeLogButton } from './ui/ViewChangeLogButton';
+import { ChangeLogService } from '../service/Updates.service';
```
byliuyang:
```suggestion
import { ChangeLogService } from '../service/ChangeLog.service';
```

byliuyang: Aha, this should change correspondingly, right?

frontend/src/component/pages/Home.tsx
```diff
@@ -91,6 +93,15 @@ export class Home extends Component<Props, State> {
     });
     this.handleStateChange();
     this.autoFillLongLink();
+
+    const changeLog = await this.props.changeLogService.getChangeLog();
+    this.setState({ changeLog }, async () => {
+      if (!(await this.props.changeLogService.hasUpdates())) {
```
byliuyang:
```suggestion
      const hasUpdates = await this.props.changeLogService.hasUpdates();
      if (!hasUpdates) {
```

frontend/src/service/ChangeLog.service.ts
```diff
@@ -0,0 +1,64 @@
...
+      const changeLog = await this.getChangeLog();
+      const lastSeenTimestamp = await this.getLastSeenChangeLog();
+      if (
+        changeLog &&
+        changeLog[0] &&
+        lastSeenTimestamp < changeLog[0].releasedAt
+      ) {
+        resolve(true);
+      }
+
```
byliuyang:
```suggestion
if (! changeLog) {
    resolve(false);
    return;
}

if (changeLog.length < 1) {
    resolve(false);
    return;
}

if (! lastSeenAt < changeLog[0].releasedAt) {
    resolve(false);
    return;
}
```

frontend/src/service/ChangeLog.service.ts
```diff
@@ -0,0 +1,64 @@
+import { Update } from '../entity/Update';
+
+export class ChangeLogService {
+  hasUpdates(): Promise<boolean> {
+    return new Promise(async (resolve, reject) => {
+      const changeLog = await this.getChangeLog();
+      const lastSeenTimestamp = await this.getLastSeenChangeLog();
```
byliuyang:
```suggestion
      const lastSeenAt = await this.getLastSeenChangeLogAt();
```

frontend/src/service/ChangeLog.service.ts
```diff
@@ -0,0 +1,64 @@
...
+        lastSeenTimestamp < changeLog[0].releasedAt
+      ) {
+        resolve(true);
+      }
+
+      resolve(false);
+    });
+  }
+
+  getLastSeenChangeLog(): Promise<number> {
```
byliuyang:
```suggestion
  getLastSeenChangeLog(): Promise<Time> {
```

byliuyang: Let's define a new type call Time to clarify what this is.

frontend/src/service/ChangeLog.service.ts
```diff
@@ -0,0 +1,73 @@
+import { Update } from '../entity/Update';
+
+type Time = number;
+
+export class ChangeLogService {
+  hasUpdates(): Promise<boolean> {
+    return new Promise(async (resolve, reject) => {
+      const changeLog = await this.getChangeLog();
+      const lastSeenAt = await this.getLastSeenChangeLog();
+      if (!changeLog) {
```
byliuyang: Move this before line 9. `getLastSeenChangeLog` is a network request and takes time. If there is no changelog, then there is no point of loading `getLastSeenChangeLog`. This also reduces our server load.

kapillamba4: @byliuyang, getChangeLog is also a network request. Most probably changeLog will not be empty so i don't think moving [these](https://github.com/short-d/short/pull/560/files#diff-b7dccee6e4430bdac2b4f2d5ce242846R10-R18) lines will not have must intended effect

PR #583, opened 2020-03-19
Copy generated link to user's clipboard automatically

### LINKED ISSUE #527: [Feature] Allow generated link to be copied into user's clipboard
**What is the problem?**
Every time user wants to generate a short link and share it with their fellows, they have to select the link text and then copy.

**Your solution**
Instead, the generated the link can be automatically copied to clipboard.

**Alternatives considered**
Instead, there can also be a button which on clicking will copy the generated link to clipboard.

**Additional context**
Instead of the current message "You can now paste `<link>` in your browser to visit.", we can have something like "Please visit `https:/short-d.com/r/<alias>` or `s/<alias>`(if you have short-d chrome extension) in your browser". Allowing this text to be copied will also publicize the existence of such an extension. Your thoughts?

rohithbalaji123:
Fixes #527 
### Description
If the short link creation is successful, the generated link is copied to user's clipboard automatically and shows a toast to notify the behavior.
### Screenshots
image

### DIFF
--- a/frontend/src/component/UIFactory.tsx
+++ b/frontend/src/component/UIFactory.tsx
@@ -19,10 +19,12 @@ import { SearchService } from '../service/Search.service';
 import { SearchBar } from './ui/SearchBar';
 import { ViewChangeLogButton } from './ui/ViewChangeLogButton';
 import { ChangeLogService } from '../service/ChangeLog.service';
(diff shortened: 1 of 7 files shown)


rohithbalaji123:
This PR is split into #597 , #598 and #603 .

PR #597, opened 2020-03-22
Add toast component to show notifications

rohithbalaji123:
### Description
Displays a toast notification at the bottom of the page. This was designed to solve #527 where the requirement is to show a toast message in the bottom of the page. So, this component is currently limited to solve that purpose. Later if required for other purposes, we can generalize it further by adding a container to show multiple toast messages at the same time and also enable customization to choose location of the toasts. Eg: top-left, bottom-right, etc.,
### Screenshots
image


frontend/src/component/ui/Toast.scss
```diff
@@ -0,0 +1,33 @@
...
+
+  @include default-font;
+}
+
+.toast {
+  z-index: 10000;
+  width: 80vw;
+  max-width: 300px;
+  color: #dedede;
+  background-color: $color-dark-gray;
```
byliuyang:
```suggestion
  background-color: $color-toast-background;
```
`$color-toast-background` can then reference `$color-dark-gray`.
The goal is to allow us changing the background color globally when retheming.

frontend/src/component/ui/Toast.scss
```diff
@@ -0,0 +1,33 @@
...
+  position: fixed;
+  opacity: 0.9;
+  bottom: 5vh;
+  left: 50%;
+  transform: translateX(-50%);
```
byliuyang: Let's add 
```
@mixin align-horizontal-center {
    left: 50%;
   transform: translateX(-50%);
}
```
to `frontend/src/component/styles/alignment` and use the mixin here to encourage reuse.

frontend/src/component/ui/Toast.spec.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { Toast } from './Toast';
+
+describe('toast', () => {
```
byliuyang:
```suggestion
describe('Toast component', () => {
```

frontend/src/component/ui/Toast.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    jest.advanceTimersByTime(1000);
+    expect(container.textContent).toContain('Toast message.');
+
+    jest.advanceTimersByTime(1000);
+    expect(container.textContent).not.toContain('Toast message.');
+
+    jest.clearAllTimers();
+  });
+
+  test('second showAndHide call should not affect first toast', () => {
```
byliuyang: I think this one should actually effect the first toast when we have two notifications overlapping. What are your thoughts?

rohithbalaji123: Ohkay, do we need to extend the delay for closing the toast or show multiple toast messages?

byliuyang:
Hi @rohithbalaji123 , thank you so much for build this cool component! I left few comments on the PR.

frontend/src/component/ui/Toast.spec.tsx
byliuyang: @rohithbalaji123 How about overriding the first Toast message and time out? I am not sure what's the best approach.

rohithbalaji123: @byliuyang Yeah, showing multiple toast messages doesn't sound right as the clipboard is overwritten by the new url.

frontend/src/component/ui/Toast.tsx
```diff
@@ -0,0 +1,67 @@
...
+  private preemptiveHide() {
+    if (this.delayTimeoutHandle)
+      clearTimeout(this.delayTimeoutHandle);
+    this.hide();
+  }
+
+  showAndHide(delay: number) {
+    // reset timer if the toast is already being shown
+    if (this.state.isShown) {
+      this.preemptiveHide();
```
byliuyang:
```suggestion
      this.cancelHide();
```

frontend/src/component/ui/Toast.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    jest.advanceTimersByTime(1500);
+    // time is currently 2500ms
+    expect(container.textContent).toContain('Toast message.');
+
+    jest.advanceTimersByTime(1000);
+    // time is currently 3500ms
+    expect(container.textContent).not.toContain('Toast message.');
+
+    jest.clearAllTimers();
```
byliuyang:
```suggestion
    const toastRef = React.createRef<Toast>();
    let toastMessage = 'Toast Message';
    const { container } = render(
      <Toast ref={toastRef} toastMessage={toastMessage} />
    );

    expect(container.textContent).not.toContain('Toast message');
    toastRef.current!.notify(2000);

    jest.advanceTimersByTime(1000);
    
    toastMessage = 'New message';
    // second notify before the first one closes(at 1000ms)
    toastRef.current!.notify(2000);
    expect(container.textContent).toContain('New message');

    jest.advanceTimersByTime(1500);
    // time is currently 2500ms
    expect(container.textContent).toContain('New message');

    jest.advanceTimersByTime(1000);
    // time is currently 3500ms
    expect(container.textContent).not.toContain('New message');

    jest.clearAllTimers();
```

frontend/src/component/ui/Toast.tsx
```diff
@@ -0,0 +1,61 @@
...
+  constructor(props: IProps) {
+    super(props);
+    this.state = {
+      isShown: false
+    };
+    this.hideTimeoutHandle = null;
+  }
+
+  public notify(duration: number) {
+    this.hideIfAlreadyShown();
```
byliuyang:
```suggestion
    this.resetTimer();
```

frontend/src/component/ui/Toast.tsx
```diff
@@ -0,0 +1,61 @@
...
+  private hideAfter(duration: number) {
+    this.hideTimeoutHandle = setTimeout(() => this.hide(), duration);
+  }
+
+  private hideIfAlreadyShown() {
+    if (this.state.isShown) {
+      this.hideTimeoutHandle && clearTimeout(this.hideTimeoutHandle);
+      this.hide();
+    }
+  }
```
byliuyang:
```suggestion
  private resetTimer() {
      if (!this.hideTimeoutHandle) {
        return;
      }
      clearTimeout(this.hideTimeoutHandle);
  }
```

frontend/src/component/ui/Toast.spec.tsx
rohithbalaji123: @byliuyang, I don't think this will work. The component is created before changing the toast message for the second time.

byliuyang: @rohithbalaji123 Yes, doesn't seem to work. It seems that we need this: https://testing-library.com/docs/example-update-props

PR #598, opened 2020-03-22
Add clipboard service to copy text to user's clipboard

rohithbalaji123:
### Description
Clipboard service enabling JavaScript to copy text into user's clipboard.


frontend/src/service/utilities/Clipboard.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    textArea.value = text;
+
+    document.body.appendChild(textArea);
+    textArea.focus();
+    textArea.select();
+
+    let successful = false;
+    try {
+      successful = document.execCommand('copy');
+    } catch (err) {}
```
byliuyang:
```suggestion
    } catch (err) {
         console.log(`Failed to copy ${text} into Clipboard`);
    }
```

frontend/src/service/utilities/Clipboard.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    return successful;
+  }
+
+  public copyTextToClipboard(text: string): Promise<void> {
+    if (!navigator.clipboard) {
+      console.log('other');
+      return this.fallbackCopyTextToClipboard(text)
+        ? Promise.resolve()
+        : Promise.reject();
+    }
```
byliuyang:
```suggestion
    if (navigator.clipboard) {
      return navigator.clipboard.writeText(text);
    }
    const isSuccessful =  this.fallbackCopyTextToClipboard(text);
    if (isSuccessful) {
        return Promise.resolve();
    }
    return Promise.reject();
```

frontend/src/service/utilities/Clipboard.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    textArea.style.width = '2em';
+    textArea.style.height = '2em';
+
+    textArea.style.padding = '0';
+    textArea.style.border = 'none';
+    textArea.style.outline = 'none';
+    textArea.style.boxShadow = 'none';
+    textArea.style.background = 'transparent';
+
+    textArea.value = text;
```
byliuyang: How about doing this instead?
textArea.style.display = 'none';

rohithbalaji123: As per this method, the component should be there in the page to be able to select the text and copy to the clipboard. So, we are adding the component to the page, selecting, copying it and removing it from the page. Setting display to none won't let us select the text inside the textArea. The answer [here](https://stackoverflow.com/questions/400212/how-do-i-copy-to-the-clipboard-in-javascript) explains it in detail.

rohithbalaji123: I tested changing the style to `textArea.style.display = 'none';` as well. It didn't work.

frontend/src/service/utilities/Clipboard.service.ts
```diff
@@ -0,0 +1,46 @@
...
+    textArea.style.left = '0';
+
+    textArea.style.width = '2em';
+    textArea.style.height = '2em';
+
+    textArea.style.padding = '0';
+    textArea.style.border = 'none';
+    textArea.style.outline = 'none';
+    textArea.style.boxShadow = 'none';
+    textArea.style.background = 'transparent';
```
byliuyang: Can this be wrapped within a private function called `createInvisibleTextArea()`?

rohithbalaji123: Okay, this looks good.

frontend/src/service/clipboardService/Clipboard.factory.ts
```diff
@@ -0,0 +1,12 @@
+import { IClipboardService } from './Clipboard.service';
+import { NativeClipboardService } from './NativeClipboard.service';
+import { SimulatedClipboardService } from './SimulatedClipboard.service';
+
+export class ClipboardFactory {
```
byliuyang:
```suggestion
export class ClipboardServiceFactory {
```

byliuyang: This file can be renamed to `Clipboard.service.factory.ts`

frontend/src/service/clipboardService/Clipboard.factory.ts
```diff
@@ -0,0 +1,12 @@
...
+  static createBrowserExtensionService(): IClipboardService {
```
byliuyang:
```suggestion
 public makeClipboardService(): IClipboardService {
```

byliuyang: We want to be able to swap out `ClipboardServiceFactory` implementation during testing using dependency injection. So we can't have static function here.

frontend/src/service/clipboardService/SimulatedClipboard.service.ts
```diff
@@ -0,0 +1,61 @@
+import { IClipboardService } from './Clipboard.service';
+
+export class SimulatedClipboardService implements IClipboardService {
+  private setupTextAreaWithText(text: string): HTMLTextAreaElement {
```
byliuyang:
```suggestion
  private addInvisibleTextArea(text: string): HTMLTextAreaElement {
```

frontend/src/service/clipboardService/SimulatedClipboard.service.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    return textArea;
+  }
+
+  private selectTextInTextArea(textArea: HTMLTextAreaElement): void {
+    textArea.focus();
+    textArea.select();
+  }
+
+  private tearDownTextArea(textArea: HTMLTextAreaElement): void {
```
byliuyang:
```suggestion
  private removeTextArea(textArea: HTMLTextAreaElement): void {
```

PR #610, opened 2020-03-29
Add resolvers for ChangeLog

kapillamba4:


backend/app/adapter/graphql/resolver/authquery.go
```diff
@@ -35,9 +37,17 @@ func (v AuthQuery) URL(args *URLArgs) (*URL, error) {
 	return &URL{url: u}, nil
 }
 
-func newAuthQuery(user *entity.User, urlRetriever url.Retriever) AuthQuery {
+// ChangeLog retrieves full ChangeLog from persistent storage
+func (v AuthQuery) ChangeLog() (ChangeLog, error) {
+	changeLog, err := v.changeLog.GetChangeLog()
+	currentTime := time.Now()
```
alldroll: It is better to use our own `timer`. For example for tests purposes

backend/app/adapter/graphql/resolver/change.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// SummaryMarkdown retrieves SummaryMarkdown of Change entity
+func (c Change) SummaryMarkdown() *string {
+	return c.change.SummaryMarkdown
+}
+
+// ReleasedAt retrieves ReleasedAt of Change entity
+func (c Change) ReleasedAt() scalar.Time {
+	return scalar.Time{Time: *c.change.ReleasedAt}
```
alldroll: Are you sure we don't have here `nil pointer exception`?

backend/app/adapter/graphql/resolver/changelog.go
```diff
@@ -0,0 +1,33 @@
...
+	lastViewedAt time.Time
+}
+
+// Changes retrieves full ChangeLog
+func (c ChangeLog) Changes() []Change {
+	return c.changeLog
+}
+
+// LastViewedAt retrieves lastViewedAt for given user
+func (c ChangeLog) LastViewedAt() *scalar.Time {
```
alldroll: Why do we need to have a pointer? If it is not crucial, we can avoid memory escape
https://medium.com/faun/golang-escape-analysis-reduce-pressure-on-gc-6bde1891d625

kapillamba4: according to schema it is 
`lastViewedAt: Time`
instead of
`lastViewedAt: Time!`
Therefore the resolver function should return a pointer which can be have nil value.

backend/app/adapter/graphql/resolver/authquery.go
kapillamba4: Yea, you are right. But we will be fetching lastViewedAt from storage in the next PR. This is just a temporary value for lastViewedAt..i think we should just set it to null for now instead of setting it to a temporary value
`return newChangeLog(changeLog, nil), err`

backend/app/adapter/graphql/resolver/changelog.go
byliuyang: @alldroll This is the limitation of the GraphQL framework we are using. It forces pointer when a value can be optional. I think the memory has to be allocated onto the heap.

backend/app/adapter/graphql/resolver/authquery.go
kapillamba4: ok, we will use mdtimer.NewTimer() for now

byliuyang: @kapillamba4 Aha, we actually mean `fw.timer`, which is an interface. `mdtimer.Timer` is concrete implementation. We can't replace it on the fly.

```golang
type AuthQuery struct {
	user         entity.User
        timer       fw.timer
	changeLog    changelog.ChangeLog
	urlRetriever url.Retriever
}

func (a AuthQuery) ChangeLog() (ChangeLog, error) {
	changeLog, err := v.changeLog.GetChangeLog()
	currentTime := a. timer.Now()
	return newChangeLog(changeLog, currentTime), err
}
```

backend/app/adapter/graphql/resolver/change.go
kapillamba4: schema is this:
```
type Change {
  	id: String!
  	title: String
  	summaryMarkdown: String
  	releasedAt: Time!
}
```

entity is this:
```

type Change struct {
	ID              string
	Title           string
	SummaryMarkdown *string
	ReleasedAt      *time.Time
}
```

should i change entity to match the schema:
```
type Change struct {
	ID              string
	Title           *string
	SummaryMarkdown *string
	ReleasedAt      time.Time
}
```

kapillamba4: Also according to schema title can be nil, that should not happen right?

byliuyang: @kapillamba4 
```golang
func (c Change) ReleasedAt() scalar.Time {
         if c.change.ReleasedAt == nil {
             return nil
         }
	return scalar.Time{Time: *c.change.ReleasedAt}
}
```
`*nil` will crash the service.

byliuyang: @kapillamba4 Yes. `Title` can't be empty.

kapillamba4: this gives error "cannot use nil as type scalar.Time in return argument"

> @kapillamba4
> 
> ```go
> func (c Change) ReleasedAt() scalar.Time {
>          if c.change.ReleasedAt == nil {
>              return nil
>          }
> 	return scalar.Time{Time: *c.change.ReleasedAt}
> }
> ```
> 
> `*nil` will crash the service.

alldroll: You can change the function signature for this case:

```go
func (c Change) ReleasedAt() scalar.Time
```

As you are not sure that `c.change.ReleasedAt` is always a valid pointer

kapillamba4: then we would have to change the schema to:

```
type Change {
  	id: String!
  	title: String
  	summaryMarkdown: String
  	releasedAt: Time
}

```

kapillamba4: updated

backend/app/adapter/graphql/resolver/authmutation.go
```diff
@@ -28,6 +30,17 @@ type CreateURLArgs struct {
 	IsPublic bool
 }
 
+// CreateChangeArgs represents the possible parameters for CreateChange endpoint
+type CreateChangeArgs struct {
+	Change *ChangeInput
```
byliuyang:
```suggestion
	Change ChangeInput
```

backend/app/adapter/graphql/resolver/authmutation.go
```diff
@@ -59,9 +72,16 @@ func (a AuthMutation) CreateURL(args *CreateURLArgs) (*URL, error) {
 	}
 }
 
-func newAuthMutation(user *entity.User, urlCreator url.Creator) AuthMutation {
+// CreateChange creates a new Change in the ChangeLog
+func (a AuthMutation) CreateChange(args *CreateChangeArgs) (Change, error) {
+	change, err := a.changeLog.CreateChange(*args.Change.Title, args.Change.SummaryMarkdown)
```
byliuyang:
```suggestion
	change, err := a.changeLog.CreateChange(args.Change.Title, args.Change.SummaryMarkdown)
```
`*` means `deference`. Dereferencing pointer is dangerous because the process will crash if the pointer is `nil`.

backend/app/adapter/graphql/resolver/authmutation.go
```diff
@@ -59,9 +72,16 @@ func (a AuthMutation) CreateURL(args *CreateURLArgs) (*URL, error) {
 	}
 }
 
-func newAuthMutation(user *entity.User, urlCreator url.Creator) AuthMutation {
+// CreateChange creates a new Change in the ChangeLog
```
byliuyang:
```suggestion
// CreateChange creates a Change in the change log
```

backend/app/adapter/graphql/resolver/authquery.go
```diff
@@ -3,15 +3,20 @@ package resolver
 import (
 	"time"
 
+	"github.com/short-d/app/fw"
+
```
byliuyang:
```suggestion
```
Let's run `./scripts/format` to reorder the imports.

backend/app/adapter/graphql/resolver/authquery.go
```diff
@@ -3,15 +3,20 @@ package resolver
...
 	"github.com/short-d/short/app/entity"
+	"github.com/short-d/short/app/usecase/changelog"
 	"github.com/short-d/short/app/usecase/url"
 )
 
 // AuthQuery represents GraphQL query resolver that acts differently based
 // on the identify of the user
 type AuthQuery struct {
 	user         *entity.User
+	timer        fw.Timer
```
byliuyang: Timer should be a dependency of `ChangeLog` instead of `AuthQuery`. Retrieving current time is a piece of core business logic, which we call an use case.  Moving timer into `ChangeLog` allows us to reuse `ChangeLog` use case without worrying about internal implementation. This also make `AuthQuery` easier to understand by removing the unrelated complexities.

backend/app/adapter/graphql/resolver/change.go
```diff
@@ -0,0 +1,38 @@
+package resolver
+
+import (
+	"github.com/short-d/short/app/adapter/graphql/scalar"
+	"github.com/short-d/short/app/entity"
+)
+
+// Change represents single change in ChangeLog
```
byliuyang:
```suggestion
// Change retrieves requested fields of a Change.
```

backend/app/adapter/graphql/resolver/change.go
```diff
@@ -0,0 +1,38 @@
...
+type Change struct {
+	change entity.Change
+}
+
+// ID retrieves ID of Change entity
```
byliuyang:
```suggestion
// ID retrieves the ID of Change entity.
```

backend/app/adapter/graphql/resolver/change.go
```diff
@@ -0,0 +1,38 @@
...
+func (c Change) ID() string {
+	return c.change.ID
+}
+
+// Title retrieves Title of Change entity
```
byliuyang:
```suggestion
// Title retrieves the title of Change entity.
```

backend/app/adapter/graphql/resolver/change.go
```diff
@@ -0,0 +1,38 @@
...
+func (c Change) Title() string {
+	return c.change.Title
+}
+
+// SummaryMarkdown retrieves SummaryMarkdown of Change entity
```
byliuyang:
```suggestion
// SummaryMarkdown retrieves the summary of Change entity in Markdown format.
```

backend/app/adapter/graphql/resolver/changelog.go
```diff
@@ -0,0 +1,33 @@
...
+	changeLog    []Change
+	lastViewedAt time.Time
+}
+
+// Changes retrieves full ChangeLog
+func (c ChangeLog) Changes() []Change {
+	return c.changeLog
+}
+
+// LastViewedAt retrieves lastViewedAt for given user
```
byliuyang:
```suggestion
// LastViewedAt retrieves the time the user viewed the change log.
```

backend/app/adapter/graphql/resolver/changelog.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/short-d/short/app/entity"
+)
+
+// ChangeLog represents ChangeLog entity and user Last Viewed ChangeLog time
+type ChangeLog struct {
+	changeLog    []Change
+	lastViewedAt time.Time
+}
+
+// Changes retrieves full ChangeLog
```
byliuyang:
```suggestion
// Changes retrieves full change log
```

backend/app/adapter/graphql/resolver/changelog.go
```diff
@@ -0,0 +1,33 @@
+package resolver
+
+import (
+	"time"
+
+	"github.com/short-d/short/app/adapter/graphql/scalar"
+	"github.com/short-d/short/app/entity"
+)
+
+// ChangeLog represents ChangeLog entity and user Last Viewed ChangeLog time
```
byliuyang:
```suggestion
// ChangeLog retrieves full change log and the time when the user viewed it.
```

backend/app/adapter/graphql/resolver/changelog.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// Changes retrieves full ChangeLog
+func (c ChangeLog) Changes() []Change {
+	return c.changeLog
+}
+
+// LastViewedAt retrieves lastViewedAt for given user
+func (c ChangeLog) LastViewedAt() *scalar.Time {
+	return &scalar.Time{Time: c.lastViewedAt}
```
byliuyang:
```suggestion
        if c.lastViewedAt == nil {
		return nil
	}
	return &scalar.Time{Time: c.lastViewedAt}
```

backend/app/adapter/graphql/schema.go
```diff
@@ -28,6 +42,11 @@ input URLInput {
 	expireAt: Time
 }
 
+input ChangeInput {
+  	title: String
```
byliuyang:
```suggestion
  	title: String!
```

backend/app/adapter/graphql/resolver/changelog.go
```diff
@@ -0,0 +1,33 @@
...
+	"time"
+
+	"github.com/short-d/short/app/adapter/graphql/scalar"
+	"github.com/short-d/short/app/entity"
+)
+
+// ChangeLog represents ChangeLog entity and user Last Viewed ChangeLog time
+type ChangeLog struct {
+	changeLog    []Change
+	lastViewedAt time.Time
```
byliuyang:
```suggestion
	lastViewedAt *time.Time
```

byliuyang:
Hi @kapillamba4 , thank you so much for working on this cool feature! I left few comments on the PR!

backend/app/adapter/graphql/resolver/changelog.go
kapillamba4: c.lastViewedAt cannot be nil it is not a pointer

backend/app/adapter/graphql/resolver/authquery.go
kapillamba4: But getting lastViewedAt should be part of use case of userChangeLog which we have not implemented yet

byliuyang: @kapillamba4 Then let’s add a fake API to changelog use case and make it work in the next PR!

backend/app/adapter/graphql/graphql_test.go
```diff
@@ -43,6 +43,6 @@ func TestGraphQlAPI(t *testing.T) {
 
 	logger := mdtest.NewLoggerFake(mdtest.FakeLoggerArgs{})
 	tracer := mdtest.NewTracerFake()
-	graphqlAPI := NewShort(&logger, &tracer, retriever, creator, verifier, authenticator)
+	graphqlAPI := NewShort(&logger, &tracer, retriever, creator, nil, verifier, authenticator)
```
byliuyang, line 46: Let's initialize `changelog.ChangeLog` with Fake or dummy dependencies instead of passing `nil`. This would cause crashes if we change the test to actually call GraphQL APIs in the future.

backend/app/adapter/graphql/resolver/authmutation.go
```diff
@@ -28,6 +30,17 @@ type CreateURLArgs struct {
 	IsPublic bool
 }
 
+// CreateChangeArgs represents the possible parameters for CreateChange endpoint
+type CreateChangeArgs struct {
+	Change ChangeInput
+}
+
+// ChangeInput represents possible Change attributes
```
byliuyang, line 38:
```suggestion
// ChangeInput represents possible properties for Change
```

backend/app/adapter/graphql/resolver/authquery.go
```diff
@@ -35,9 +37,16 @@ func (v AuthQuery) URL(args *URLArgs) (*URL, error) {
 	return &URL{url: u}, nil
 }
 
-func newAuthQuery(user *entity.User, urlRetriever url.Retriever) AuthQuery {
+// ChangeLog retrieves full ChangeLog from persistent storage
+func (v AuthQuery) ChangeLog() (ChangeLog, error) {
+	changeLog, err := v.changeLog.GetChangeLog()
+	return newChangeLog(changeLog, v.changeLog.GetLastViewedAt()), err
```
byliuyang, line 43:
```suggestion
	changeLog, err := v.changeLog.GetChangeLog()
	lastViewedAt := v.changeLog.GetLastViewedAt()
	return newChangeLog(changeLog, lastViewedAt), err
```

backend/app/adapter/graphql/resolver/authquery_test.go
```diff
@@ -86,7 +86,7 @@ func TestAuthQuery_URL(t *testing.T) {
 
 			fakeRepo := repository.NewURLFake(testCase.urls)
 			retrieverFake := url.NewRetrieverPersist(&fakeRepo)
-			query := newAuthQuery(&testCase.user, retrieverFake)
+			query := newAuthQuery(&testCase.user, nil, retrieverFake)
```
byliuyang, line 89: Let's initialize and pass in real changeLog instead of `nil`.

backend/app/adapter/graphql/resolver/query_test.go
```diff
@@ -60,7 +60,7 @@ func TestQuery_AuthQuery(t *testing.T) {
 			retrieverFake := url.NewRetrieverPersist(&fakeRepo)
 			logger := mdtest.NewLoggerFake(mdtest.FakeLoggerArgs{})
 			tracer := mdtest.NewTracerFake()
-			query := newQuery(&logger, &tracer, authenticator, retrieverFake)
+			query := newQuery(&logger, &tracer, authenticator, nil, retrieverFake)
```
byliuyang, line 63: Let's initialize real change log here.

backend/app/adapter/graphql/schema.go
```diff
@@ -16,10 +16,24 @@ type Mutation {
...
+type ChangeLog {
+  	changes: [Change!]!
+  	lastViewedAt: Time
+}
+
+type Change {
+  	id: String!
+  	title: String!
+  	summaryMarkdown: String
+  	releasedAt: Time
```
byliuyang, line 31:
```suggestion
  	releasedAt: Time!
```
It's seem that every change we published should require `releasedAt`. What are your thoughts on this?

backend/app/usecase/changelog/changelog.go
```diff
@@ -43,6 +46,12 @@ func (p Persist) GetChangeLog() ([]entity.Change, error) {
 	return p.changeLogRepo.GetChangeLog()
 }
 
+// GetLastViewedAt retrieves the last time the user viewed the change log
+func (p Persist) GetLastViewedAt() *time.Time {
+	now := p.timer.Now()
```
byliuyang, line 51: Let's add a TODO here so we can implement this feature in the next PR!
TODO format: 
```
// TODO(issue#your_issue_number): fetch the last time the user viewed the change log from persistent storage.
```

backend/app/adapter/db/changelog.go
```diff
@@ -40,7 +40,7 @@ FROM "%s";`,
 		if err != nil {
 			return changeLog, err
 		}
-		change.ReleasedAt = utc(change.ReleasedAt)
+		change.ReleasedAt = *utc(&change.ReleasedAt)
```
byliuyang, line 43:
```suggestion
		change.ReleasedAt = change.ReleasedAt.UTC()
```
The utc helper was created to conveniently handle `nil` value. `change.ReleasedAt` can't be `nil` here. So that's not a problem.

backend/app/adapter/graphql/resolver/authquery_test.go
```diff
@@ -7,7 +7,12 @@ import (
 	"time"
 
 	"github.com/short-d/app/mdtest"
+	"github.com/short-d/short/app/adapter/db"
 	"github.com/short-d/short/app/adapter/graphql/scalar"
+	"github.com/short-d/short/app/usecase/changelog"
+	"github.com/short-d/short/app/usecase/keygen"
+	"github.com/short-d/short/app/usecase/service"
+
```
byliuyang, line 15:
```suggestion
```

backend/app/adapter/graphql/resolver/query_test.go
```diff
@@ -6,6 +6,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/short-d/short/app/adapter/db"
+	"github.com/short-d/short/app/usecase/changelog"
+	"github.com/short-d/short/app/usecase/keygen"
+	"github.com/short-d/short/app/usecase/service"
+
```
byliuyang, line 13:
```suggestion
```

backend/app/adapter/graphql/resolver/authquery_test.go
kapillamba4: i think i ran ./script/format this time

Issue #640, opened 2020-04-08
[Refactor] Replace hard-coded auth token valid duration with value from env variable


byliuyang:
Currently, auth token valid duration is hard-coded in [wire.go](https://github.com/short-d/short/blob/master/backend/dep/wire.go). This value should be extracted from environmental variable to avoid unnecessary code changes in the future.

arberiii:
@byliuyang Can I work on this issue?

PR #645, opened 2020-04-08
Replace hard-coded auth token valid duration with value from env variable

arberiii:
Fix [issue 640](https://github.com/short-d/short/issues/640)


backend/dep/wire.go
```diff
@@ -36,14 +37,12 @@ import (
...
-// TODO(issue#640): replace with value from env variable.
-const oneDay = 24 * time.Hour
-const oneWeek = 7 * oneDay
+var tokenDuration = getTokenDuration()
 
 var authSet = wire.NewSet(
 	provider.NewJwtGo,
 
-	wire.Value(provider.TokenValidDuration(oneWeek)),
+	wire.Value(provider.TokenValidDuration(tokenDuration)),
```
alldroll, line 45: We should stick the one way how to pass configurable values. Please take a look at how we do it `https://github.com/short-d/short/blob/master/backend/main.go#L54`

backend/dep/wire.go
```diff
@@ -211,3 +210,22 @@ func InjectRoutingService(
 	)
 	return mdservice.Service{}
 }
+
+func getEnv(key, defaultValue string) string {
```
alldroll, line 214: These functions already exist https://github.com/short-d/short/blob/master/backend/main.go#L85
We should reduce the number of code duplicates as much as we can.

backend/.env.dist
```diff
@@ -21,4 +21,6 @@ JWT_SECRET=random
 WEB_FRONTEND_URL=http://localhost:3000
 KEY_GEN_BUFFER_SIZE=10
 KEY_GEN_HOSTNAME=kgs1-staging.short-d.com
-KEY_GEN_PORT=443
\ No newline at end of file
+KEY_GEN_PORT=443
+
+AUTH_TOKEN_DURATION=604800000000
```
alldroll, line 26: This number doesn't tell me anything. Is it seconds, minutes? It is better to add a brief comment or reflect it in the variable's name `AUTH_TOKEN_DURATION_IN_SEC`

arberiii: Thank you for the comment, I change it to AUTH_TOKEN_LIFETIME_IN_SEC!

backend/dep/wire.go
arberiii: I removed the mustInt function from backend/main and placed an exportable function in dep package. But for getEnv I'm not sure how to do it without creating cycles because we need AUTH_TOKEN_LIFETIME_IN_SEC inside dep package

alldroll: I meant to move initialization to `main.go` and add the corresponding field to the `Config` struct

byliuyang:
Hi @arberiii , thank you so much for working on this! 
I should be more clear on the issue. This refactor requires you to change 
```go
var authSet = wire.NewSet(
	provider.NewJwtGo,

	wire.Value(provider.TokenValidDuration(oneWeek)),
	provider.NewAuthenticator,
)
```
to
```go
func authSet(tokenValidDuration provider.TokenValidDuration) wire.ProviderSet {
    return wire.NewSet(
	provider.NewJwtGo,

	wire.Value(tokenValidDuration),
	provider.NewAuthenticator,
    )
}
func InjectRoutingService(
...
tokenValidDuration provider.TokenValidDuration
) {
    wire.Build(
		wire.Bind(new(fw.StdOut), new(mdio.StdOut)),
		...
		wire.Bind(new(fw.GraphQlRequest), new(mdrequest.GraphQL)),

		observabilitySet,
		authSet(tokenValidDuration),
}
```

Ideally, we want to pass `5m`, `2h`, `5d`, `1w`, ... so that the config is highly readable like the following one:
Screen Shot 2020-04-08 at 9 07 59 AM

Hi @alldroll, @arberiii is new to the team. Next time, when people are confused, please provide an example to clear communicate your expectations.

byliuyang:
I'll implement the struct thing @alldroll mentioned to help the next refactor.

backend/main.go
```diff
@@ -14,7 +12,7 @@ func main() {
 
 	host := env.GetEnv("DB_HOST", "localhost")
 	portStr := env.GetEnv("DB_PORT", "5432")
-	port := mustInt(portStr)
+	port := dep.MustInt(portStr)
```
byliuyang, line 15: `dep` package is only for configuring dependencies and not for helpers.

arberiii:
> ```go
> InjectRoutingService
> ```

I modified in that way but when I run wire command then it returns unknown pattern at the line 
> ```go
> authSet(tokenValidDuration),
> ```

PR #648, opened 2020-04-08
Rename GetURLs to GetURLsByUser in url usecase

### LINKED ISSUE #647: [Refactor] Rename GetURLs method in url use case to GetURLsByUser
**What is the problem?**
GraphQL resolver will pass user object instead of a pointer to url usecase

**Your solution**
Rename GetURLs method in url usecase to GetURLsByUser

kapillamba4:
Fixes #647

### DIFF
--- a/backend/app/usecase/url/urlretriever.go
+++ b/backend/app/usecase/url/urlretriever.go
@@ -1,7 +1,6 @@
 package url
 
 import (
-	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


backend/app/usecase/url/urlretriever.go
```diff
@@ -58,15 +57,8 @@ func (r RetrieverPersist) getURL(alias string) (entity.URL, error) {
 	return url, nil
 }
 
-// GetURLs retrieves URLs created by given user from persistent storage
-func (r RetrieverPersist) GetURLs(user *entity.User) ([]entity.URL, error) {
-	if user == nil {
-		return nil, errors.New("fail to find the owner of the URLs")
-	}
-	return r.getURLsByUser(*user)
```
byliuyang, line 66: I wonder what happens to this?

kapillamba4: viewer returns object of type entity.user now (https://github.com/short-d/short/pull/638/files#diff-7f09a888a7007a4d8968cb85f5d8d2beR10) which will be used to fetch user details in AuthQuery Resolver, whether user is nil or not (authToken is valid or not) is handled internally in the auth usecase.

byliuyang: @kapillamba4 Cool!

PR #650, opened 2020-04-09
Automate env config parsing with reflection

byliuyang:
### Description
All the configs are manual retrieved from environmental variables and converted into the correct data type. This added significant amount of unrelated code into `main.go`, making it difficult for developers to understand the existing configs or add new configs.

### Description
Use `reflection` to automatically retrieve environmental variables and parse them into a predefined struct. `main.go` now looks much cleaner.

### DIFF
--- a/backend/envconfig/envconfig.go
+++ b/backend/envconfig/envconfig.go
@@ -20,8 +20,12 @@ type EnvConfig struct {
 // parse them into the given struct.
 func (e EnvConfig) ParseConfigFromEnv(config interface{}) error {
 	configVal := reflect.ValueOf(config)
-	if configVal.Kind() != reflect.Ptr || configVal.IsNil() {
+	if configVal.Kind() != reflect.Ptr {
 		return errors.New("config must be a pointer")
+	}
+
+	if configVal.IsNil() {
+		return errors.New("config can't be nil")
 	}
 
 	elem := configVal.Elem()
--- a/backend/envconfig/envconfig.go
+++ b/backend/envconfig/envconfig.go
@@ -1,5 +1,5 @@
-// TODO(issue#649): move this package into app framework
 package envconfig
+// TODO(issue#649): move this package into app framework
 
 import (
 	"errors"
@@ -10,10 +10,13 @@ import (
 	"github.com/short-d/app/fw"
 )
 
+// EnvConfig parses configuration from environmental variables.
 type EnvConfig struct {
 	environment fw.Environment
 }
 
+// ParseConfigFromEnv retrieves configurations from environmental variables and
+// parse them into the given struct.
 func (e EnvConfig) ParseConfigFromEnv(config interface{}) error {
 	configVal := reflect.ValueOf(config)
 	if configVal.Kind() != reflect.Ptr || configVal.IsNil() {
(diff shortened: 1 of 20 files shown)


backend/envconfig/envconfig.go
```diff
@@ -0,0 +1,79 @@
...
+		fieldValue.SetInt(int64(num))
+		return nil
+	case reflect.Bool:
+		boolean, err := strconv.ParseBool(newValue)
+		if err != nil {
+			return err
+		}
+		fieldValue.SetBool(boolean)
+		return nil
+	default:
```
byliuyang, line 71: @arberiii You can add a case for time to support parsing human readable string, such as '1week', into time.

byliuyang: https://github.com/short-d/short/pull/645

byliuyang: @alldroll What are your thoughts?

alldroll: I think we can reflect duration in naming. The code should be as simple as possible. And `1week` is going to work only for 1 case:)

You did a great job! Let's move this to `app` repository in the next PR

backend/envconfig/envconfig.go
```diff
@@ -0,0 +1,79 @@
...
+type EnvConfig struct {
+	environment fw.Environment
+}
+
+// ParseConfigFromEnv retrieves configurations from environmental variables and
+// parse them into the given struct.
+func (e EnvConfig) ParseConfigFromEnv(config interface{}) error {
+	configVal := reflect.ValueOf(config)
+	if configVal.Kind() != reflect.Ptr || configVal.IsNil() {
+		return errors.New("config must be a pointer")
```
arberiii, line 24: Should the error message also include that the maybe the config is nil?

backend/envconfig/envconfig.go
arberiii: This would definitely help with PR [#645](https://github.com/short-d/short/pull/645)!

backend/envconfig/envconfig.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil
+	case reflect.Bool:
+		boolean, err := strconv.ParseBool(newValue)
+		if err != nil {
+			return err
+		}
+		fieldValue.SetBool(boolean)
+		return nil
+	default:
+		return fmt.Errorf("unexpected field type: %s", kind)
```
arberiii, line 72: What about other types such as uint8, uint16, etc.? I think that there is no field other than string and int in ServiceConfig struct, is that the reason?

backend/envconfig/envconfig.go
byliuyang: @alldroll naming? I think we should define human readable formats for common units and create unit package to help manage that. CacheSize int ‘env: “CACHE_SIZE” unit:”storage”’  Degrees int ‘env: “DEGREES” unit:”temperature”’

arberiii: This would definitely help with PR [#645](https://github.com/short-d/short/pull/645)!

backend/envconfig/envconfig.go
byliuyang: @arberiii Sure! We can separate the checks!

backend/envconfig/envconfig.go
byliuyang: @arberiii We’ll support more and more as needed! It doesn’t make sense for us to add everything at the moment ( we also have other priorities )

PR #653, opened 2020-04-10
Replace hard-coded auth token valid duration with value from env variable

arberiii:
Fix [issue 640](https://github.com/short-d/short/issues/640)


backend/envconfig/envconfig.go
```diff
@@ -72,6 +74,14 @@ func setFieldValue(field reflect.StructField, fieldValue reflect.Value, newValue
 		}
 		fieldValue.SetBool(boolean)
 		return nil
+	case reflect.Int64:
+		// right now the only env var of type int64 is AuthTokenLifetime
```
arberiii, line 78: there is no reflect.Kind of type time.Duration

byliuyang: Screen Shot 2020-04-10 at 9 35 17 AM

byliuyang: I guess we will have a function called process `int`. Inside which, we will do some `if else`.

backend/unit/unit.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// ParseDuration reads a human-readable duration and returns duration in seconds
+func ParseDuration(s string) (time.Duration, error) {
+	var (
+		duration time.Duration
+		err      error
+		length   int
+	)
+	timeVariable := s[len(s)-1]
```
byliuyang, line 21:
```suggestion
	unit := s[len(s)-1]
```

backend/unit/unit.go
```diff
@@ -0,0 +1,38 @@
...
+	length, err = strconv.Atoi(s[:len(s)-1])
```
byliuyang, line 22:
```suggestion
	value, err = strconv.Atoi(s[:len(s)-1])
```

backend/unit/unit.go
```diff
@@ -0,0 +1,38 @@
...
+	"time"
+)
+
+const (
+	oneDay  = time.Hour * 24
+	oneWeak = oneDay * 7
+)
+
+// ParseDuration reads a human-readable duration and returns duration in seconds
+func ParseDuration(s string) (time.Duration, error) {
```
byliuyang, line 15:
```suggestion
func ParseDuration(readableDuration string) (time.Duration, error) {
```

backend/unit/unit.go
```diff
@@ -0,0 +1,38 @@
...
+	var (
+		duration time.Duration
+		err      error
+		length   int
+	)
```
byliuyang, line 20: Lazy define variables at the place we need them.

backend/unit/unit.go
```diff
@@ -0,0 +1,38 @@
+package unit
```
byliuyang, line 1: Love this package!

backend/envconfig/envconfig.go
byliuyang: Please check the value of `Type.PkgPath` and `Type.Name`

backend/unit/unit.go
```diff
@@ -13,24 +13,30 @@ const (
...
+		err      error
+		length   int
+	)
 	timeVariable := s[len(s)-1]
 	switch timeVariable {
 	case 's':
-		length, err := strconv.Atoi(s[:len(s)-1])
-		return time.Duration(length) * time.Second, err
+		length, err = strconv.Atoi(s[:len(s)-1])
+		duration = time.Duration(length) * time.Second
```
byliuyang: I like direct returns. It reduces complexities.

backend/unit/unit.go
```diff
@@ -0,0 +1,38 @@
...
+
+// ParseDuration reads a human-readable duration and returns duration in seconds
+func ParseDuration(readableDuration string) (time.Duration, error) {
+	var (
+		duration time.Duration
+		err      error
+		value    int
+	)
+	unit := readableDuration[len(readableDuration)-1]
+	value, err = strconv.Atoi(readableDuration[:len(readableDuration)-1])
```
byliuyang, line 22: Err check and early return here.

backend/cmd/cmd.go
```diff
@@ -29,6 +30,7 @@ type ServiceConfig struct {
 	KeyGenBufferSize     int
 	KgsHostname          string
 	KgsPort              int
+	AuthTokenLifetime    time.Duration
```
byliuyang, line 33: FYI, in the future, we may want to consider eliminating this repetitive struct.

backend/envconfig/envconfig.go
```diff
@@ -8,6 +8,8 @@ import (
 	"reflect"
 	"strconv"
 
+	"github.com/short-d/short/unit"
+
```
byliuyang, line 12:
```suggestion
```

backend/envconfig/envconfig.go
```diff
@@ -81,3 +83,14 @@ func setFieldValue(field reflect.StructField, fieldValue reflect.Value, newValue
 func NewEnvConfig(environment fw.Environment) EnvConfig {
 	return EnvConfig{environment: environment}
 }
+
+func processInt(value, kind string) (int64, error) {
```
byliuyang, line 87:
```suggestion
func processInt(newValue, type reflect.Type) (int64, error) {
```
I think we need to ensure this duration is the official one define inside built-in `time` package.
`PkgPath` will help with this.

backend/envconfig/envconfig.go
```diff
@@ -81,3 +83,14 @@ func setFieldValue(field reflect.StructField, fieldValue reflect.Value, newValue
...
+	if kind == "int" {
+		num, err := strconv.Atoi(value)
+		return int64(num), err
+	} else if kind == "Duration" {
```
byliuyang, line 91:
```suggestion
	}
	if pkg == "time" && kind == "Duration" {
```

backend/envconfig/envconfig.go
```diff
@@ -81,3 +83,14 @@ func setFieldValue(field reflect.StructField, fieldValue reflect.Value, newValue
 func NewEnvConfig(environment fw.Environment) EnvConfig {
 	return EnvConfig{environment: environment}
 }
+
+func processInt(value, kind string) (int64, error) {
+	if kind == "int" {
```
byliuyang, line 88: nit: maybe in the future we can consider switch statement.

backend/envconfig/envconfig.go
byliuyang: Re thought about this. How about renaming to `parseInt`? `ProcessInt` didn't provide reader enough information.

backend/unit/unit.go
```diff
@@ -0,0 +1,40 @@
...
+	"strconv"
+	"time"
+)
+
+const (
+	oneDay  = time.Hour * 24
+	oneWeak = oneDay * 7
+)
+
+// ParseDuration reads a human-readable duration and returns duration in seconds
```
byliuyang, line 14: I guess we won't support `ms` for now. How about adding a TODO so that we can think about extending to `ms` or adding other units, such as storage (B, KB, MB, GB, TB, PB) and `temperature ( C, F )`. This also implies that we may include other data structures for `temperature` or `storage` representation.

backend/unit/unit_test.go
```diff
@@ -0,0 +1,182 @@
...
+			expectedDuration time.Duration
+		}{
+			{
+				name:             "regular seconds",
+				time:             "3s",
+				expectedHasError: false,
+				expectedDuration: 3 * time.Second,
+			},
+			{
+				name:             "irregular seconds",
```
byliuyang, line 27:
```suggestion
				name:             "incorrect format",
```

backend/unit/unit_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+	t.Run("second", func(t *testing.T) {
+		testCases := []struct {
+			name             string
+			time             string
+			expectedHasError bool
+			expectedDuration time.Duration
+		}{
+			{
+				name:             "regular seconds",
```
byliuyang, line 21:
```suggestion
				name:             "correct format",
```

backend/unit/unit_test.go
```diff
@@ -0,0 +1,182 @@
...
+	t.Parallel()
+
+	t.Run("second", func(t *testing.T) {
+		testCases := []struct {
+			name             string
+			time             string
+			expectedHasError bool
+			expectedDuration time.Duration
+		}{
+			{
```
byliuyang, line 20: We should also add cases for `empty string` and `multi digits` parsing ( 10s, 120s )

byliuyang: We should also add a case for custom define `Duration` struct to make sure it fails because it's not the duration from the built-in `time` package.

backend/.env.dist
```diff
@@ -21,4 +21,6 @@ JWT_SECRET=random
 WEB_FRONTEND_URL=http://localhost:3000
 KEY_GEN_BUFFER_SIZE=10
 KEY_GEN_HOSTNAME=kgs1-staging.short-d.com
-KEY_GEN_PORT=443
\ No newline at end of file
+KEY_GEN_PORT=443
+
+AUTH_TOKEN_LIFETIME=1w
```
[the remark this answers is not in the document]
arberiii: Thank you 😊

backend/cmd/cmd.go
arberiii: I was thinking it too, the config struct is becoming too big to manually initialize!

backend/envconfig/envconfig.go
arberiii: True, parseInt gives more insights.

backend/envconfig/envconfig.go
arberiii: I'll change it now!

arberiii:
I will split this PR into little PRs

PR #655, opened 2020-04-11
Create unit package

arberiii:
### Description
Have a central place to parse units.

### DIFF
--- a/backend/unit/unit.go
+++ b/backend/unit/unit.go
@@ -1,0 +1,41 @@
+package unit
+
+import (
+	"errors"
+	"strconv"
+	"time"
+)
+
+const (
+	oneDay  = time.Hour * 24
+	oneWeak = oneDay * 7
+)
+
+// ParseDuration reads a human-readable duration and returns duration in seconds
+// TODO: extend to ms or add other units, such as storage (B, KB, MB, GB, TB, PB) and temperature ( C, F).
+func ParseDuration(readableDuration string) (time.Duration, error) {
+	if len(readableDuration) == 0 {
+		return 0, errors.New("duration string is empty")
+	}
+
+	value, err := strconv.Atoi(readableDuration[:len(readableDuration)-1])
+	if err != nil {
+		return 0, err
+	}
+
+	unit := readableDuration[len(readableDuration)-1]
+	switch unit {
+	case 's':
+		return time.Duration(value) * time.Second, nil
+	case 'm':
+		return time.Duration(value) * time.Minute, nil
+	case 'h':
+		return time.Duration(value) * time.Hour, nil
+	case 'd':
+		return time.Duration(value) * oneDay, nil
+	case 'w':
+		return time.Duration(value) * oneWeak, nil
+	default:
+		return 0, errors.New("unknown time type")
+	}
+}
--- a/backend/unit/unit.go
+++ b/backend/unit/unit.go
@@ -7,7 +7,6 @@ import (
 )
 
 // TODO(issue#659): extend and add other units
-
 const (
 	oneDay  = time.Hour * 24
 	oneWeak = oneDay * 7
--- a/backend/unit/unit.go
+++ b/backend/unit/unit.go
@@ -6,24 +6,27 @@ import (
 	"time"
 )
 
+// TODO(issue#659): extend and add other units
+
 const (
 	oneDay  = time.Hour * 24
 	oneWeak = oneDay * 7
 )
 
-// ParseDuration reads a human-readable duration and returns duration in seconds
-// TODO: extend to ms or add other units, such as storage (B, KB, MB, GB, TB, PB) and temperature ( C, F).
+// ParseDuration converts a human-readable duration and returns duration into seconds
 func ParseDuration(readableDuration string) (time.Duration, error) {
 	if len(readableDuration) == 0 {
-		return 0, errors.New("duration string is empty")
+		return 0, errors.New("readable duration can't be empty")
 	}
 
-	value, err := strconv.Atoi(readableDuration[:len(readableDuration)-1])
+	end := len(readableDuration) - 1
+	valueStr := readableDuration[:end]
+	value, err := strconv.Atoi(valueStr)
 	if err != nil {
 		return 0, err
 	}
 
-	unit := readableDuration[len(readableDuration)-1]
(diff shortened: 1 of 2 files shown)


backend/unit/unit.go
```diff
@@ -0,0 +1,41 @@
...
+	"time"
+)
+
+const (
+	oneDay  = time.Hour * 24
+	oneWeak = oneDay * 7
+)
+
+// ParseDuration reads a human-readable duration and returns duration in seconds
+// TODO: extend to ms or add other units, such as storage (B, KB, MB, GB, TB, PB) and temperature ( C, F).
```
byliuyang, line 15: Here is our TODO format: TODO format: // TODO(issue#631): support responsive layouts.

backend/unit/unit.go
```diff
@@ -0,0 +1,41 @@
...
+	"strconv"
+	"time"
+)
+
+const (
+	oneDay  = time.Hour * 24
+	oneWeak = oneDay * 7
+)
+
+// ParseDuration reads a human-readable duration and returns duration in seconds
```
byliuyang, line 14:
```suggestion
// ParseDuration converts a human-readable duration and returns duration into seconds
```

backend/unit/unit.go
```diff
@@ -0,0 +1,41 @@
...
+// TODO: extend to ms or add other units, such as storage (B, KB, MB, GB, TB, PB) and temperature ( C, F).
+func ParseDuration(readableDuration string) (time.Duration, error) {
+	if len(readableDuration) == 0 {
+		return 0, errors.New("duration string is empty")
```
byliuyang, line 18:
```suggestion
		return 0, errors.New("readable duration can't be empty")
```

backend/unit/unit.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	value, err := strconv.Atoi(readableDuration[:len(readableDuration)-1])
```
byliuyang, line 21:
```suggestion
        end := len(readableDuration) - 1
	valueStr  := readableDuration[:end]
	value, err := strconv.Atoi(valueStr)
```

backend/unit/unit.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	unit := readableDuration[len(readableDuration)-1]
```
byliuyang, line 26:
```suggestion
	unit := readableDuration[end]
```

backend/unit/unit.go
```diff
@@ -0,0 +1,41 @@
...
+	case 'm':
+		return time.Duration(value) * time.Minute, nil
+	case 'h':
+		return time.Duration(value) * time.Hour, nil
+	case 'd':
+		return time.Duration(value) * oneDay, nil
+	case 'w':
+		return time.Duration(value) * oneWeak, nil
+	default:
+		return 0, errors.New("unknown time type")
```
byliuyang, line 39:
```suggestion
		return 0, errors.New("unknown duration unit")
```

backend/unit/unit_test.go
```diff
@@ -0,0 +1,207 @@
...
+			fakeDuration     fakeDuration
+		}{
+			{
+				name:             "correct format",
+				time:             "3s",
+				expectedHasError: false,
+				expectedDuration: 3 * time.Second,
+			},
+			{
+				name:             "multi digits",
```
byliuyang, line 31:
```suggestion
				name:             "multiple digits",
```

backend/unit/unit_test.go
```diff
@@ -0,0 +1,207 @@
...
+				time:             "3ss",
+				expectedHasError: true,
+				expectedDuration: 0,
+			},
+			{
+				name:         "custom Duration type",
+				time:         "6s",
+				hasFakeType:  true,
+				fakeDuration: fakeDuration(6 * time.Second),
+			},
```
byliuyang, line 53: I think this should be part of the envconfig. The solo responsibility of unit is to parse unit, not dealing with reflection

backend/unit/unit_test.go
```diff
@@ -0,0 +1,207 @@
...
+		for _, testCase := range testCases {
+			t.Run(testCase.name, func(t *testing.T) {
+				duration, err := ParseDuration(testCase.time)
+				if testCase.expectedHasError {
+					mdtest.NotEqual(t, nil, err)
+					return
+				} else if testCase.hasFakeType {
+					mdtest.NotEqual(t, testCase.fakeDuration, duration)
+					return
+				}
```
byliuyang, line 65:
```suggestion
				}
```

backend/unit/unit_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"testing"
+	"time"
+
+	"github.com/short-d/app/mdtest"
+)
+
+type fakeDuration int64
+
+func TestParseDuration(t *testing.T) {
+	t.Parallel()
```
byliuyang, line 13: Let's add empty string and multiple digits case for each unit.

arberiii: Empty string is handled the same way for each unit so it's not an added value for tests, however we can add test when there units but no value i.e. "m", "w"

byliuyang: @arberiii When writing unit tests, we should know as less internal implementation as possible. What happens if we changed the code. Then the test won't fail. The goal of the test is to make sure we are safe from refactors.

backend/unit/unit_test.go
arberiii: Should this be as part of other PR because the parsing of duration is still not added to this PR?

backend/unit/unit_test.go
arberiii: Rethinking about it, that is a really valuable lesson. Thank you!

backend/unit/unit.go
```diff
@@ -0,0 +1,44 @@
+package unit
+
+import (
+	"errors"
+	"strconv"
+	"time"
+)
+
+// TODO(issue#659): extend and add other units
+
```
byliuyang, line 10:
```suggestion
```

backend/unit/unit_test.go
byliuyang: @arberiii Yes, this should be part of the other PR!

Issue #193, opened 2019-10-01
Add React component to show all short links created by the user


byliuyang:
Currently, users can't see all the short links they created.

<img width="1299" alt="Screen Shot 2019-09-30 at 10 37 51 PM" src="https://user-images.githubusercontent.com/3537801/65936901-f84ddb00-e3d2-11e9-93b3-aca552a69a67.png">

Add React component with dummy data to show all the linked created by the user after he, she or they signed in.

shade34321:
I'd like to try this if this is still available but I'll probably have lots of questions.

byliuyang:
@shade34321 Sounds good. Please feel free to ask question or suggest refactor.

byliuyang:
@shade34321 Please join this [Slack channel](https://s.time4hacks.com/r/short-slack) to discuss bugs, dev environment setup, tooling, and coding best practices.

shade34321:
Oh man, I never saw this message. I"ll see if I can work on it later this weekend or next. I'm sorry!

byliuyang:
The component should be hidden by feature toggle by default.

JackieYJC:
I can take this task. Seems like a very important feature to the user!

JackieYJC:
For this task, does the UI need to be in place? In other words, is the expectation that after it is done, whoever works on the backend API for read/write to the database can just populate the component and enable the feature to live?

byliuyang:
@JackieYJC Mostly yes. For now, we just populate fake data and turn off the feature toggle. We will have to make small change to our frontend data service to make sure it pulls data from real backend.

JackieYJC:
> @JackieYJC Mostly yes. For now, we just populate fake data and turn off the feature toggle. We will have to make small change to our frontend data service to make sure it pulls data from real backend.

Gotcha, I'm considering using tables from this source: https://material-ui.com/components/tables/

byliuyang:
@JackieYJC Prefer building the components ourselves so that we:
1) easily customize the styles
2) add new features
3) learn how the internals work

JackieYJC:
@byliuyang Yes, I don't intend to copy-and-paste anything. Just learn their tricks and apply it here. A table is a table 😛

byliuyang:
@JackieYJC Let's split this issue into few PRs:

1) The reusable table component
2) The reusable pagination component
3) The `ShortLinks`  component
4) Fake  `ShortLinkService`
[...]

PR #674, opened 2020-04-18
Add ShortLink Sevice

rohithbalaji123:
Part of #193 

### Description
- Added a separate ApiService class to take care of all API calls to consolidate Graph related libraries in a single class.
- Added ShortLinkService with functionality to invoke querying user created short links.

### Screenshots
image


frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,92 @@
...
+    authQuery(authToken: $authToken) {
+      URLs {
+        alias
+        originalURL
+      }
+    }
+  }
+`;
+
+export class GraphQLApiService implements IApiService {
```
byliuyang, line 35: I think we should separate `GraphQLApiService` and `IApiService`.  There will be REST ( feature toggle ) and WebSocket APIs.

byliuyang: `IApiService` seems unnecessary.

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,92 @@
...
+      return [Err.NetworkError];
+    }
+    if (!graphQLErrors || graphQLErrors.length === 0) {
+      return [Err.Unknown];
+    }
+    return graphQLErrors.map(this.gqlErrorToCode);
+  }
+
+  private gqlErrorToCode(graphQLError: IGraphQLError): string {
+    if (graphQLError.message === 'token expired') {
```
byliuyang, line 87: We should compare code instead of raw message. Message can be easily changed. Code won't. https://github.com/short-d/short/blob/master/frontend/src/service/GraphQL.service.ts#L18

frontend/src/service/ShortLink.service.ts
```diff
@@ -1,6 +1,51 @@
...
+            return;
+          }
+          reject({
+            getUserShortLinksErr: this.errorService.getErr(errCode)
+          });
+        });
+    });
+  }
+
+  private getPagedShortLinksFromURLs(
```
byliuyang, line 40:
```suggestion
  private getPagedShortLinks(
```

frontend/src/service/ShortLink.service.ts
```diff
@@ -1,6 +1,51 @@
...
+  getUserCreatedShortLinks(
+    offset: number,
+    pageSize: number
+  ): Promise<IPagedShortLinks> {
+    return new Promise((resolve, reject) => {
+      this.apiService
+        .invokeGetUserShortLinksApi(offset, pageSize)
+        .then(URLs => {
+          resolve(this.getPagedShortLinksFromURLs(URLs, offset, pageSize));
+        })
```
byliuyang, line 27: I think the TODO should be added here instead.

frontend/src/service/Api.service.ts
rohithbalaji123: Yeah, the issue being, query doesn't return extension code. This is the error response template for graphQL queries.

byliuyang: Interesting. It works for alias already exist error.

byliuyang: @rohithbalaji123 Let's return https://github.com/short-d/short/blob/70f3eef11f3fe5a0f27546f970b02ef7019c69c6/backend/app/adapter/graphql/resolver/error.go#L13 at https://github.com/short-d/short/blob/70f3eef11f3fe5a0f27546f970b02ef7019c69c6/backend/app/adapter/graphql/resolver/authquery.go#L52

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
+import { AuthService } from './Auth.service';
+import { Err, ErrorService } from './Error.service';
+import { EnvService } from './Env.service';
+import { GraphQLService, IGraphQLError } from './GraphQL.service';
+import { Url } from '../entity/Url';
+
+interface IGraphQLSchemaQuery {
```
byliuyang, line 7:
```suggestion
interface IShortGraphQLQuery {
```

byliuyang: We may call GraphQL apis by other providers. So it's important to make it clear that this one is for Short.

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
...
+interface IGraphQLSchemaAuthQuery {
+  URLs?: IGraphQLSchemaURL[];
+}
+
+interface IGraphQLSchemaURL {
+  alias: string;
+  originalURL: string;
+}
+
+const GET_USER_SHORT_LINKS_QUERY = `
```
byliuyang, line 20: Let's embedded this into the `invokeGetUserShortLinksApi` and rename the function name to `getUserShortLinks`. I don't think this query will be reused in other method of `ApiService`.

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
...
+    authQuery(authToken: $authToken) {
+      URLs {
+        alias
+        originalURL
+      }
+    }
+  }
+`;
+
+export class ApiService {
```
byliuyang, line 31:
```suggestion
export class ShortGraphQLApiService {
```

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
...
+  private graphQLBaseURL: string;
```
byliuyang, line 32:
```suggestion
  private baseURL: string;
```

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
...
+    private errorService: ErrorService,
+    private envService: EnvService,
+    private graphQLService: GraphQLService
+  ) {
+    this.graphQLBaseURL = `${this.envService.getVal(
+      'GRAPHQL_API_BASE_URL'
+    )}/graphql`;
+  }
+
+  invokeGetUserShortLinksApi(offset: number, pageSize: number): Promise<Url[]> {
```
byliuyang, line 45:
```suggestion
  getUserShortLinks(offset: number, pageSize: number): Promise<Url[]> {
```

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
...
+    return new Promise((resolve, reject) => {
+      this.graphQLService
+        .query<IGraphQLSchemaQuery>(this.graphQLBaseURL, {
+          query: GET_USER_SHORT_LINKS_QUERY,
+          variables: { authToken: this.authService.getAuthToken() }
```
byliuyang, line 50:
```suggestion
          variables: { authToken: this.authService.getAuthToken() }
```
Let's move this to a separate variable on top of line 46.

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
...
+import { EnvService } from './Env.service';
+import { GraphQLService, IGraphQLError } from './GraphQL.service';
+import { Url } from '../entity/Url';
+
+interface IGraphQLSchemaQuery {
+  authQuery: IGraphQLSchemaAuthQuery;
+}
+
+interface IGraphQLSchemaAuthQuery {
+  URLs?: IGraphQLSchemaURL[];
```
byliuyang, line 12: I think the `?` after `URL` is not needed.

Here is the API definition for AuthQuery: https://github.com/short-d/short/blob/44a1208e6f2e5b08ff42294d3f88e236f025cf42/backend/app/adapter/graphql/schema.go#L20

From the API definition, `URLs` is not nullable.

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
...
+  invokeGetUserShortLinksApi(offset: number, pageSize: number): Promise<Url[]> {
+    return new Promise((resolve, reject) => {
+      this.graphQLService
+        .query<IGraphQLSchemaQuery>(this.graphQLBaseURL, {
+          query: GET_USER_SHORT_LINKS_QUERY,
+          variables: { authToken: this.authService.getAuthToken() }
+        })
+        .then((res: IGraphQLSchemaQuery) => {
+          const { URLs } = res.authQuery;
+          resolve(URLs!.map(this.getUrlFromGraphQLApiURL));
```
byliuyang, line 54:
```suggestion
          resolve(URLs.map(this.parseUrl));
```

frontend/src/service/Api.service.ts
```diff
@@ -0,0 +1,88 @@
...
+        })
+        .catch(err => {
+          const errCodes = this.getGraphQLQueryErrorCodes(err);
```
byliuyang, line 57:
```suggestion
          const errCodes = this.getErrorCodes(err);
```

frontend/src/service/Api.service.ts
byliuyang: We probably need a switch case here.

kapillamba4: Shall i create a PR for this part @byliuyang?

kapillamba4: Why a switch case? what other error types are possible here?

byliuyang: @kapillamba4 For now we are just handling unauthenticated error. We may also handle other types of errors separately.

byliuyang: @kapillamba4 If you want to open an PR, then it's great!

rohithbalaji123: Nice @kapillamba4, thanks for the fix. That saved a lot of time.

byliuyang: @rohithbalaji123 He fixed a different package. It's going to take 10 mins to fix.

frontend/src/service/Api.service.ts
rohithbalaji123: @byliuyang , please do correct me if I am wrong.
Doesn't this depend on the fetch query we have? If I am querying only for URLs, is it expected to return other members like ChangeLog?

byliuyang: @rohithbalaji123 You will only get URLs.

frontend/src/service/ShortGraphQLApiService.ts
```diff
@@ -0,0 +1,91 @@
...
+          query: getUserShortLinksQuery,
+          variables: variables
+        })
+        .then((res: IShortGraphQLQuery) => {
+          const { URLs } = res.authQuery;
+          resolve(URLs.map(this.parseUrl));
+        })
+        .catch((err: IGraphQLRequestError) => {
+          const errCodes = this.getErrorCodes(err);
+          reject(errCodes[0]);
```
byliuyang, line 60: Nit: I guess this assumes `errCodes` can't have length of `0`.

rohithbalaji123: Yeah, the method makes sure it returns array containing at least one element.

PR #683, opened 2020-04-22
Return authentication error in change log API

kapillamba4:
Part of #193


backend/app/adapter/graphql/resolver/authquery.go
```diff
@@ -40,16 +41,21 @@ func (v AuthQuery) URL(args *URLArgs) (*URL, error) {
 
 // ChangeLog retrieves full ChangeLog from persistent storage
 func (v AuthQuery) ChangeLog() (ChangeLog, error) {
+	user, err := viewer(v.authToken, v.authenticator)
```
byliuyang, line 44:
```suggestion
	_, err := viewer(v.authToken, v.authenticator)
```

Let's do this for now to avoid having to change the interface for `GetLastViewedAt`.

Issue #613, opened 2020-04-01
[Feature] Implement lastViewedAt for ChangeLog use case


kapillamba4:
**What is the problem?**
Last viewed change log time for every user is currently set to a temporary value when retrieved using GraphQL API query.

**Your solution**
Allow fetching last time the user viewed the change log from persistent storage with help of UserChangeLog usecase.

kapillamba4:
Schema for user_change_log:
```

CREATE TABLE "user_change_log"
(
    "user_id"              CHARACTER VARYING(5),
    "email"                CHARACTER VARYING(254) NOT NULL PRIMARY KEY,
    "last_viewed_at"       TIMESTAMP WITH TIME ZONE
);

```

byliuyang:
`user_changelog_relation`

kapillamba4:
Steps:
1. Create migration script for `user_changelog`
2. Create a repository for UserChangeLog with methods `GetLastViewedAt` and `UpdateLastViewedAt`.
3. Create UserChangeLogSQL struct in db package with methods `GetLastViewedAt` and `UpdateLastViewedAt`.
4. Write integration tests for UserChangeLogSQL.
5. Write UserChangeLog usecase.
6. Add UserChangeLogFake in the repository.
7. Write Unit Tests for UserChangeLog use case.
8. Write GraphQL resolvers and update schema.

byliuyang:
Here are some of the steps I came up:
1. Create SQL migration for `user_changelog` table.
2. Create `UserChangeLog` Repository with the following methods ( integration tests included ):
-  `GetLastViewedAt`
-  `UpdateLastViewedAt`
3. Add `ViewChangeLog` method to `ChangeLog` use case and update `GetLastViewedAt` ( may have to pass in `User`). No fake implementation is need. Must include `unit tests`. 
4. Update `GraphQL` scheme and add corresponding `resolvers`.

PR #684, opened 2020-04-22
Pass user as a parameter to GetLastViewedAt

kapillamba4:
Part of #613


byliuyang:
Can we implement the use case first so that we are sure this change is the correct approach?

kapillamba4:
errors.As does not work as expected @byliuyang, we would need to use `switch err.(type) {...}`, otherwise it works fine

byliuyang:
@kapillamba4 Can you private message some logs to look up tomorrow?

PR #686, opened 2020-04-23
Add feature toggle API

byliuyang:
### Description
Feature toggle status are hard coded at the frontend. Enabling or disabling a feature requires developers to make code changes.

### Description
Create feature toggle API so that we can dynamically turn on or off a feature.


backend/dep/wire.go
```diff
@@ -5,6 +5,8 @@ package dep
 import (
 	"database/sql"
 
+	"github.com/short-d/short/app/usecase/feature"
```
arberiii, line 8: Is this formated?

byliuyang: An empty line will cause issues...

backend/app/adapter/routing/handle.go
```diff
@@ -79,3 +81,19 @@ func NewSSOSignInCallback(
...
+	instrumentationFactory request.InstrumentationFactory,
+	featureDecisionFactory feature.DecisionFactory,
+) fw.Handle {
+	return func(w http.ResponseWriter, r *http.Request, params fw.Params) {
+		i := instrumentationFactory.NewHTTP(r)
+		featureID := params["featureID"]
+
+		decision := featureDecisionFactory.NewDecision(i)
+		isEnable := decision.IsFeatureEnable(featureID)
+		body, _ := json.Marshal(isEnable)
```
arberiii, line 96: I guess that in case of error the body will be nil therefore there is no need to check for errors, am I right?

byliuyang: @arberiii I think I was just being lazy. Should add instrumentation here.

PR #717, opened 2020-04-29
Add col names prop to Table & add col names to UserShortLinksSection

### LINKED ISSUE #702: [Refactor] [UI] Fix column widths in Favorites section
**What is frustrating you?**
When traversing between pages, the column widths in favourites section changes depending on the length on long links making the alias column keep moving.

**Your solution**
Allot a percentage of table widths for each column to make sure the column remain in the same position irrespective of the length of the long links.

columnWidth

Coteh:
Fixes #702.

### Description
Column widths in "Favorites" section is not consistent. If a *long* long link is shown on a cell within the left column, it will stretch the whole column out for that page.
### Screenshots
See PR #702.

### Description
Allocated 70% of the width to the long link column and 30% of the width to the alias column (only one of the CSS rules are needed but I added both to be explicit). This will fix the problem.
### Screenshots
result

## TODO
- [x] Write new test(s) for `Table`
- [x] ~Write new test(s) for `UserShortLinksSection`~ decided not to write tests here as the only major change to this component was style-related


frontend/src/component/pages/shared/UserShortLinksSection.scss
```diff
@@ -14,4 +14,12 @@
...
     word-break: break-all;
   }
+
+  .table-cell.long {
+    width: 70%;
+  }
+  
+  .table-cell.alias {
+    width: 30%;
+  }
```
byliuyang, line 24: Let's remove this for now.

frontend/src/component/ui/Table.tsx
```diff
@@ -1,22 +1,24 @@
 import React, { Component, ReactChild } from 'react';
 
 import './Table.scss';
+import classNames from 'classnames';
 
 interface IProps {
   headers?: ReactChild[];
   rows?: ReactChild[][];
+  colNames?: string[];
```
byliuyang, line 9:
```suggestion
  colClassNames?: string[];
```

frontend/src/component/ui/Table.tsx
```diff
@@ -1,22 +1,24 @@
...
+  private createHeaders(headers: ReactChild[] | undefined, colNames: string[] | undefined) {
     if (!headers || headers.length === 0) {
       return null;
     }
     return (
       <tr key={`header`}>
         {headers.map((cell: ReactChild, cellIndex: number) => {
           return (
-            <th key={`cell-${cellIndex}`} className="table-cell">
+            <th key={`cell-${cellIndex}`} className={!colNames ? "table-cell" : classNames("table-cell", colNames[cellIndex])}>
```
byliuyang, line 21: Let's simplify this one. We want to remove table-cell and keep column class Names only.

rohithbalaji123:
@Coteh , can you please check the case for long alias as well?

alias_bug

frontend/src/component/ui/Table.tsx
Coteh: My next commit will contain this fix.

Coteh:
Now accommodate for long aliases.

result

frontend/src/component/ui/Table.tsx
```diff
@@ -1,22 +1,24 @@
 import React, { Component, ReactChild } from 'react';
 
 import './Table.scss';
+import classNames from 'classnames';
```
Coteh, line 4: Remove this import

frontend/src/component/pages/shared/UserShortLinksSection.scss
```diff
@@ -11,7 +11,13 @@
...
   }
 
-  .long-link {
+  .long-link,
+  .alias {
     word-break: break-all;
   }
+
+  td.cell-long,
+  th.cell-long {
```
byliuyang, line 20:
```suggestion
  .cell-long {
```

Coteh:
Spent a bit of time playing around with `table-layout: auto` and `table-layout: fixed`. Although both options seemed to be unaffected by the table content after the `word-break: all` property is set, I ultimately went with `table-layout: fixed` as [it guarantees that the table layout will be unaffected by table contents](https://css-tricks.com/fixing-tables-long-strings/). Also see [this](https://css-tricks.com/almanac/properties/t/table-layout/) too: _"The column values used are based on widths defined on columns or cells for the first row of the table."_

Additionally, I added some tests to `Table` for the col class names. However, I couldn't find a way to test column width using `react-testing-library` (column sizes appear to be 0 when rendered using the test framework), and I believe this is by design as sizes of elements rendered is more of a browser implementation detail and I don't believe this testing library is designed for testing implementation details in this manner. So therefore I just have the two tests in this PR.

Ready for review.

Coteh:
I noticed that `prettier` didn’t seem to like my TypeScript conditional operator. However, after upgrading `prettier` to the latest version (1.19.1) it went away
```
src/component/ui/Table.tsx
[error] src/component/ui/Table.tsx: SyntaxError: Expression expected. (20:68)
[error]   18 |         {headers.map((cell: ReactChild, cellIndex: number) => {
[error]   19 |           return (
[error] > 20 |             <th key={`cell-${cellIndex}`} className={colClassNames?.[cellIndex]}>
[error]      |                                                                    ^
[error]   21 |               {cell}
[error]   22 |             </th>
[error]   23 |           );
```
it looked like this

@byliuyang Should we proceed with upgrading `prettier` in this PR?

frontend/src/component/ui/Table.tsx
Coteh: Done

Coteh:
Ready for review

frontend/src/component/pages/shared/UserShortLinksSection.scss
Coteh: Already done, but GitHub still thinks this is a change to be requested for some reason.

frontend/src/component/ui/Table.spec.tsx
```diff
@@ -129,4 +129,60 @@ describe('Table component', () => {
...
+
+    const tableHeaders = container.querySelectorAll('thead > tr > th');
+    for (let column = 0; column < sampleHeaders.length; column++) {
+      expect(tableHeaders[column].className).toEqual('');
+    }
+
+    const tableRows = container.querySelectorAll('tbody > tr');
+    for (let row = 0; row < sampleRows.length; row++) {
+      const curRowData = tableRows[row].querySelectorAll('td');
+      for (let column = 0; column < sampleRows[row].length; column++) {
```
byliuyang, line 152:
```suggestion

      for (let column = 0; column < sampleRows[row].length; column++) {
```

frontend/src/component/ui/Table.spec.tsx
```diff
@@ -129,4 +129,60 @@ describe('Table component', () => {
...
+
+    const tableHeaders = container.querySelectorAll('thead > tr > th');
+    for (let column = 0; column < sampleHeaders.length; column++) {
+      expect(tableHeaders[column].className).toEqual(sampleClassNames[column]);
+    }
+
+    const tableRows = container.querySelectorAll('tbody > tr');
+    for (let row = 0; row < sampleRows.length; row++) {
+      const curRowData = tableRows[row].querySelectorAll('td');
+      for (let column = 0; column < sampleRows[row].length; column++) {
```
byliuyang, line 183:
```suggestion

      for (let column = 0; column < sampleRows[row].length; column++) {
```

frontend/src/component/ui/Table.tsx
```diff
@@ -25,33 +32,43 @@ export class Table extends Component<IProps> {
...
 
-  private createBodyRow(row: ReactChild[]) {
+  private createBodyRow(
+    row: ReactChild[],
+    colClassNames: string[] | undefined
+  ) {
     return row.map((cell: ReactChild, cellIndex: number) => {
       return (
-        <td key={`cell-${cellIndex}`} className="table-cell">
+        <td key={`cell-${cellIndex}`} className={colClassNames?.[cellIndex]}>
```
byliuyang, line 57: Please use 
```typescript
classNames({
          colClassNames?.[cellIndex]: true
        })
```
So that we can extend the style in the future.

frontend/src/component/pages/shared/UserShortLinksSection.tsx
```diff
@@ -38,6 +38,7 @@ export class UserShortLinksSection extends Component<IProps> {
           <Table
             headers={['Long Link', 'Alias']}
             rows={this.createTableRows()}
+            colClassNames={['cell-long', 'cell-alias']}
```
byliuyang, line 41: To recommend passing in `columnWidths` instead of `colClassNames` because `colClassNames` is implementation detail of table component. The component using relays on those classes may break when we change the internal of the table.

Coteh: The `colClassNames` property was added to `Table` component so that columns can have their widths styled in via SCSS, but if we're going to pass the widths directly instead, then `colClassNames` should be deleted and instead replaced with `columnWidths`, which will add the widths directly to the columns.

byliuyang: @Coteh Please remove `colClassNames` to enforce encapsulation. Thank you!

Coteh:
Changed to widths. Should appear the same as the previous implementation, just now the widths are set explicitly to each element.

result

Issue #612, opened 2020-03-31
[Feature] Create admin dashboard for Change logs related queries


rohithbalaji123:
This issue is a start to a bigger issue #347 which concentrates on adding an admin dashboard with a constrained functionalities of supporting CRUD operations related to Change Logs.
This issue will have multiple PRs split in the form of,
- [ ] Design a wireframe for the layout of the admin page (offline, will be updated once finalized).
- [ ] Design [reusable] React components for the purpose.
- [ ] Implement the UI with dummy data.
- [ ] Add CRUD GraphQL APIs for change logs (C and R already covered in #566 ).
- [ ] Integrate the above functionalities together with actual data.

byliuyang:
@rohithbalaji123 We also need to implementation `authorizer` to handle permissions: https://martinfowler.com/articles/web-security-basics.html#AuthorizeActions

byliuyang:
For your information, I am refactoring authenticator: https://github.com/short-d/short/pull/611

rohithbalaji123:
@byliuyang 
Okay, updated the issue.
Also, I guess, we would need a separate login page for admin site. Having a modal like in main site doesn't look completely secure for admin actions. What are your thoughts?

byliuyang:
@rohithbalaji123 I saw some critical large scale system sharing the same sign in page. However, I do see them having two factor authentication. We may want to force 2 - factor authentication for admin accounts.

byliuyang:
Here is template prototype I made: https://github.com/short-d/short/issues/651
Need your help to clean it up and productionize the code.

byliuyang:
![Changelog Screen](https://user-images.githubusercontent.com/3537801/84588045-0022a380-add9-11ea-8717-1eafe1631bcf.jpg)

byliuyang:
![Changelog Screen](https://user-images.githubusercontent.com/3537801/84588095-4a0b8980-add9-11ea-9881-af702297951a.jpg)

byliuyang:
![Changelog Screen - Create Change](https://user-images.githubusercontent.com/3537801/84588581-80e39e80-addd-11ea-8b90-d561af6f2bf6.jpg)

PR #731, opened 2020-05-02
Replace material icons with custom Icon component

rohithbalaji123:
Part of #612 

### Description
Replace material icons with custom Icon component.
#### Change Logs Modal
image

#### Error Modal
image

#### Search Icon
image


frontend/src/component/ui/Icon.spec.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react';
+import { Icon, IconID } from './Icon';
+
+describe('Navigation component', () => {
```
byliuyang, line 5:
```suggestion
describe('Icon component', () => {
```

frontend/src/component/ui/Icon.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    const icon = container.querySelector('.icon');
+    expect(icon).toBeTruthy();
+
+    expect(onClickHandler).not.toHaveBeenCalled();
+    fireEvent.click(icon!);
+    expect(onClickHandler).toHaveBeenCalled();
+  });
+
+  test('should change icon when icon setter is triggered', () => {
```
byliuyang, line 36:
```suggestion
  test('should change icon when setIcon is invoked', () => {
```

frontend/src/component/ui/Icon.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+    const iconRef = React.createRef<Icon>();
+    const { container } = render(
+      <Icon ref={iconRef} defaultIconID={IconID.Search} />
+    );
+
+    const SVGBeforeChange = render(iconRef.current!.renderSVG(IconID.Search));
+    expect(container.querySelector('.icon')!.innerHTML).toContain(
+      SVGBeforeChange.container.innerHTML
+    );
```
byliuyang, line 45: I recommend assigning each icon a class and query that class instead.

frontend/src/component/ui/Icon.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    iconRef.current!.setIcon(IconID.Close);
+    const SVGAfterChange = render(iconRef.current!.renderSVG(IconID.Close));
+    expect(container.querySelector('.icon')!.innerHTML).toContain(
+      SVGAfterChange.container.innerHTML
+    );
```
byliuyang, line 51: I recommend assigning each icon a class and query that class instead.

rohithbalaji123:
@byliuyang , I have also simplified the Icon component removing the functionality of dynamically changing SVGs of the same Icon component. It seemed a bit unnatural to just change the SVG in the same Icon instance. What are your thoughts?

In AdminPage, the menu icon will be rendered as follows,
```
<Icon
  iconID={this.state.isMenuOpen ? IconID.MenuOpen : IconID.Menu}
  onClick={this.handleMenuIconClick}
/>
```

byliuyang:
@rohithbalaji123 Prefer not having this conditional. With in the same icon component allows us to do animated transitions in the future.

frontend/src/component/ui/Icon.tsx
```diff
@@ -0,0 +1,98 @@
...
+      case IconID.Close:
+        return (
+          <svg
+            className={'icon-close'}
+            xmlns="http://www.w3.org/2000/svg"
+            viewBox="0 0 24 24"
+          >
+            <path d="M0 0h24v24H0z" fill="none" />
+            <path d="M19 6.41L17.59 5 12 10.59 6.41 5 5 6.41 10.59 12 5 17.59 6.41 19 12 13.41 17.59 19 19 17.59 13.41 12z" />
+          </svg>
```
byliuyang, line 61: I rethought about this. I think we should move the `SVGs` into `private renderClose` to avoid have a super long function.

PR #740, opened 2020-05-05
Permission

### LINKED ISSUE #729: [Feature] Manage access control through authorizer
**What is the problem?**
Currently, there is no way to restrict an user from accessing protected resources, such as updating the change log, disable malicious user account, disable malicious short link, edit short links, view Analytics, or manage user roles.

**Your solution**
Create authorizer to manage user roles and permissions. For permission modal, we use RBAC: role-based access control.

**Additional context**
https://short-d.com/r/security
Sample code: https://github.com/short-d/short/tree/permission

alldroll:
It resolves #729


backend/app/adapter/db/table/userrole.go
```diff
@@ -0,0 +1,12 @@
...
+// UserRole represents database table columns for 'user_role' table
+var UserRole = struct {
+	TableName    string
+	ColumnUserID string
+	ColumnRoles  string
+}{
+	TableName:    "user_role",
+	ColumnUserID: "user_id",
+	ColumnRoles:  "roles",
+}
```
byliuyang, line 12: https://github.com/short-d/short/issues/729#issuecomment-624184699

backend/app/usecase/authorizer/authorizer.go
```diff
@@ -0,0 +1,37 @@
...
+// hasPermission tells if a user has a right to the given permission
+func (a Authorizer) hasPermission(user entity.User, permission permission.Permission) (bool, error) {
+	roles, err := a.userRoleRepo.GetUserRoles(user)
+
+	if err != nil {
+		return false, err
+	}
+
+	for _, role := range roles {
+		if role.IsAllowed(permission) {
```
byliuyang, line 27:
```suggestion
		if role.hasPermission(permission) {
```

backend/app/usecase/repository/userrole.go
```diff
@@ -0,0 +1,12 @@
+package repository
+
+import (
+	"github.com/short-d/short/app/entity"
+	"github.com/short-d/short/app/usecase/authorizer/role"
+)
+
+type UserRole interface {
+	GetUserRoles(user entity.User) ([]role.Role, error)
```
byliuyang, line 9:
```suggestion
	GetRoles(user entity.User) ([]role.Role, error)
```

backend/app/adapter/db/userrole.go
```diff
@@ -0,0 +1,213 @@
...
+
+	sort.Slice(roles, func(i, j int) bool {
+		return roles[i] > roles[j]
+	})
+
+	var encoded string
+
+	for r := roles[0]; r >= 0; r-- {
+		if len(roles) > 0 && r == roles[0] {
+			encoded += "1"
```
arberiii, line 192: Since the roles are stored in the database as array of characters, can we just append the character of the role for example if role = 2 we'd append string(2) and the decoding is done in a similar way?

alldroll: It is not going to work. How to handle the next situation:
```
RoleA = 1
RoleB = 2
...
RoleC = 21
RoleD = 22
```

Simple append doesn't work here:)

byliuyang: @arberiii We have to do shift and masking. Each bit represents whether an role is assigned an user. 1 means yes, 0 means no.

@alldroll Please correct me if I am wrong. I saw in Kernel, lots of configuration management has been done this way for storage efficiency.

arberiii: Sorry for misleading you. What I meant is for each role having a rune. Not sure what is the first character in runes, but if take
```
RoleA = 65
RoleB = 66
...
RoleC = 85
RoleD = 86
```
Then we would encode it as string(65)+string(66)+string(85)+string(86) which is equal to "ABUV". Again we are limited but I think 100+ roles are enough.

byliuyang: @arberiii We had conversation offline and switched away from bitwise design for now. @alldroll is working on a design proposal and will let us know when we can include your design in the proposal.

arberiii: I see now that it is stored as bits I thought it was as characters, my bad!

alldroll:
Is closed because of https://github.com/short-d/short/pull/742

PR #741, opened 2020-05-05
Generate random alias if empty alias provided

### LINKED ISSUE #728: [Refactor] Generate random alias if empty is passed for alias from frontend
**What is frustrating you?**
When you specify an empty string custom alias (currently impossible from the frontend but a custom form can be crafted to do this) and click "Create Short Link", it is passed as an empty string into the backend. This will end up producing a short link of empty string alias, which is undesired.

**Your solution**
The backend should interpret an empty string alias as a request to randomly generate an alias. Additionally, if malicious input is detected for the alias, also generate a random alias and log that malicious input was attempt (but do NOT provide any feedback to user that malicious input was attempted, just provide normal output)

**Additional context**
image

In the DB, if you look up the record it'll look like this

image

An alternative solution was considered to check for empty string and pass `null` as the argument instead to backend, but malicious user can easily remove this check from the frontend code.

image

(as an aside I believe there's an unknown error in above screenshot because I did not have KGS enabled locally, which is used for randomly generating aliases)

Coteh:
Fixes blank alias portion of #728. Currently just addresses empty aliases, TBD if malicious input for alias should also be handled. (edit: Will address in #744 to reduce scope of this PR)

### Description
Adding short link with alias set to empty string (not possible from frontend but GraphQL request can be crafted to have alias be set to empty string) will create entry with alias set to empty string.
### Screenshots
See #728.

### Description
Now generates a random alias whenever empty string alias is passed. It should now be impossible to have a short link with alias of empty string.

### Screenshots
image

## TODO
- [ ] Determine if other edge case aliases should be included in this PR, including malicious strings (not malicious URLs as that's already taken care of)

### DIFF
--- a/backend/app/usecase/url/urlcreator.go
+++ b/backend/app/usecase/url/urlcreator.go
@@ -68,7 +68,7 @@ func (c CreatorPersist) CreateURL(url entity.URL, customAlias *string, user enti
 		return entity.URL{}, ErrMaliciousLongLink(longLink)
 	}
 
-	if customAlias == nil || *customAlias == "" {
+	if !c.isAliasProvided(customAlias) {
 		return c.createURLWithAutoAlias(url, user)
 	}
 
@@ -76,6 +76,10 @@ func (c CreatorPersist) CreateURL(url entity.URL, customAlias *string, user enti
 		return entity.URL{}, ErrInvalidCustomAlias(*customAlias)
 	}
 	return c.createURLWithCustomAlias(url, *customAlias, user)
+}
+
+func (c CreatorPersist) isAliasProvided(customAlias *string) bool {
+	return customAlias != nil && *customAlias != ""
 }
 
 func (c CreatorPersist) createURLWithAutoAlias(url entity.URL, user entity.User) (entity.URL, error) {
--- a/backend/app/usecase/url/urlcreator.go
+++ b/backend/app/usecase/url/urlcreator.go
@@ -68,7 +68,7 @@ func (c CreatorPersist) CreateURL(url entity.URL, customAlias *string, user enti
 		return entity.URL{}, ErrMaliciousLongLink(longLink)
 	}
 
-	if customAlias == nil {
+	if customAlias == nil || *customAlias == "" {
 		return c.createURLWithAutoAlias(url, user)
 	}
 
--- a/backend/app/usecase/url/urlcreator_test.go
+++ b/backend/app/usecase/url/urlcreator_test.go
@@ -24,6 +24,7 @@ func TestURLCreatorPersist_CreateURL(t *testing.T) {
 
 	alias := "220uFicCJj"
 	longAlias := "an-alias-cannot-be-used-to-specify-default-arguments"
+	emptyAlias := ""
 
 	testCases := []struct {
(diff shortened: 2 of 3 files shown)


backend/app/usecase/url/urlcreator_test.go
```diff
@@ -93,13 +94,8 @@ func TestURLCreatorPersist_CreateURL(t *testing.T) {
 			},
 		},
 		{
-			name: "automatically generate alias",
-			urls: urlMap{
-				"220uFicCJj": entity.URL{
```
Coteh, line 98: @byliuyang I wasn't sure why this URL was in the fake in-memory URL map even though the alias wasn't being used in the test case so I took it out for now, but can add it back in if it's actually needed.

backend/app/usecase/url/urlcreator.go
```diff
@@ -68,7 +68,7 @@ func (c CreatorPersist) CreateURL(url entity.URL, customAlias *string, user enti
 		return entity.URL{}, ErrMaliciousLongLink(longLink)
 	}
 
-	if customAlias == nil {
+	if customAlias == nil || *customAlias == "" {
```
byliuyang, line 71: Please extract `customAlias == nil || *customAlias == ""` into a function called `isAliasProvided`

backend/app/usecase/url/urlcreator_test.go
```diff
@@ -24,6 +24,7 @@ func TestURLCreatorPersist_CreateURL(t *testing.T) {
 
 	alias := "220uFicCJj"
 	longAlias := "an-alias-cannot-be-used-to-specify-default-arguments"
+	emptyAlias := ""
```
byliuyang, line 27: Let's hard code empty string in the test case so that reader can easily exam it works.

backend/app/usecase/url/urlcreator_test.go
```diff
@@ -117,6 +113,26 @@ func TestURLCreatorPersist_CreateURL(t *testing.T) {
 				CreatedAt:   &utc,
 			},
 		},
+		{
+			name: "automatically generate alias if empty string alias provided",
+			urls: urlMap{},
+			availableKeys: []external.Key{
+				"test",
+			},
+			alias: &emptyAlias,
```
byliuyang, line 122: Let's hard code empty alias here.

Coteh: I did it this way because you can't pass the address of a string literal into a string pointer variable. See the following:
https://stackoverflow.com/questions/61004337/cannot-convert-untyped-string-constant-to-string
https://stackoverflow.com/questions/42594789/initialize-string-pointer-in-struct
We would have to update `CreateURL` method to accept `string` literal instead (which I believe is represented as an address anyways in Go) and `URLInput`'s `CustomAlias` field should also be updated to be `string` literal.

backend/app/usecase/url/urlcreator.go
Coteh: Done

Coteh:
Reducing scope of this PR, will continue to address other invalid/malicious forms of alias input in #744.

PR #748, opened 2020-05-07
Fix Page control index when total number of pages changes

### LINKED ISSUE #658: [BUG] Handle cases where totalPages
**Describe the bug**
The `currentPageIdx` in `PageControl` component is not updated when `totalPages` prop is updated from the parent component. There is a possibility of a `totalPages` decreasing and if `currentPageIdx` is greater than the newly passed `totalPages` prop, it causes unexpected behavior.

**To Reproduce**
This is not in production yet. To effective reproduce this, create the component in HomePage with like following,
```
<PageControl
   totalPages={this.state.totalPages!}
   onPageChanged={this.onPageChanged}
/>
```
```
onPageChanged = () => {
    if (Math.random() > 0.5) {
      this.setState({ totalPages: 10 });
    } else {
      this.setState({ totalPages: 5 });
   }
};
```
**Expected behavior**
The PageControl should automatically change it's state to currently available last page if total pages count becomes less than the current page index.

**Screenshots**
pcBug

**Additional context**
In context with our application, this case could arise when user chooses to delete the only available URL in the last page of UserShortLinksSection.

Coteh:
Fixes #658.

### Description
If page is changed to a page whose number is greater than total amount of pages (can happen when number of pages changes in between page changes), then the current page will still be set to that page instead of falling back to the new last page.

### Screenshots
See #658

### Description
Now when the total number of pages changes after a page change, the page selected will fall back to the last page.

### Screenshots
result

## Note
This solution uses `getDerivedStateFromProps`, which is [not a recommended approach to handling derived state](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html). An alternative solution would be letting the parent of `PageControl` component control the current index, but I think this will introduce unneeded complexity. Additionally, the logic within the `getDerivedStateFromProps` method will not override any local changes to the `currentPageIdx` value if it's less than or equal to the total number of pages, which is expected behaviour.

### DIFF
--- a/frontend/src/component/ui/PageControl.spec.tsx
+++ b/frontend/src/component/ui/PageControl.spec.tsx
@@ -153,9 +153,7 @@ describe('PageControl component', () => {
     fireEvent.click(pageNumbers[8]);
     expect(currentPageIdx).toBe(8);
 
-    rerender(
-      <PageControl totalPages={5} onPageChanged={handlePageChanged} />
-    );
+    rerender(<PageControl totalPages={5} onPageChanged={handlePageChanged} />);
 
     const prevButton = container.querySelector('.previous');
     expect(prevButton).toBeTruthy();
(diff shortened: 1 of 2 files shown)


byliuyang:
@Coteh This is not a bug. It' the limitation offset based pagination. I was working on a design proposal and prototype to address those. The true solution is using cursor based pagination.

Coteh:
Whoops my bad I'll close.

PR #749, opened 2020-05-08
Add tabs component

rohithbalaji123:
Part of #612 

### Description
Add Tabs component which can be handy when there a set of pages of which one should be shown at a time and rendering should be controlled with the index of the pages.
### Screenshots
image

### DIFF
--- a/frontend/src/component/ui/Tabs.spec.tsx
+++ b/frontend/src/component/ui/Tabs.spec.tsx
@@ -7,12 +7,12 @@ describe('Tabs component', () => {
     render(<Tabs />);
   });
 
-  test('should not render anything when there are no tabs', () => {
+  test('should not render anything when there is no tab', () => {
     const { container } = render(<Tabs />);
     expect(container.innerHTML).toBeFalsy();
   });
 
-  test('should render first tab content by default', () => {
+  test('should render first tab by default', () => {
     const tabs = ['Content 1', 'Content 2', 'Content 3'];
 
     const { container } = render(<Tabs>{tabs}</Tabs>);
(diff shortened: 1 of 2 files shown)


frontend/src/component/ui/Tabs.spec.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { Tabs } from './Tabs';
+
+describe('Tabs component', () => {
+  test('should render without crash', () => {
+    render(<Tabs />);
+  });
+
+  test('should not render anything when there are no tabs', () => {
```
byliuyang, line 10:
```suggestion
  test('should not render anything when there is no tab', () => {
```

frontend/src/component/ui/Tabs.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const { container } = render(<Tabs />);
+    expect(container.innerHTML).toBeFalsy();
+  });
+
+  test('should render first tab content by default', () => {
```
byliuyang, line 15:
```suggestion
  test('should render first tab by default', () => {
```

frontend/src/component/ui/Tabs.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+  test('should render correct tab content when showTab is invoked', () => {
+    const tabs = ['Content 1', 'Content 2', 'Content 3'];
+    const tabRef = React.createRef<Tabs>();
+
+    const { container } = render(<Tabs ref={tabRef}>{tabs}</Tabs>);
+
+    for (let tabIdx = 0; tabIdx < tabs.length; tabIdx++) {
+      tabRef.current!.showTab(tabIdx);
+      expect(container.textContent).toContain(tabs[tabIdx]);
+    }
```
byliuyang, line 44: I think we can replace this with a specific tab:

1) Check on another tab
2) Invoke showTab
3) Check tab changed

frontend/src/component/ui/Tabs.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  test('should not render anything when negative index is given', () => {
```
byliuyang, line 47:
```suggestion
  test('should render nothing when negative index is given', () => {
```

Issue #546, opened 2020-03-10
[BUG] Falky CI build


kapillamba4:
**Describe the bug**
CI build fails if there are concurrent backend integration tests running. 
Example: https://ci.time4hacks.com/short-d/short/1758

**Expected behavior**
CI build should be successful.

byliuyang:
Should also migrate down after test is finished. Sometimes, there are PRs having different DB scheme. This results in conflicts in db migration.

PR #752, opened 2020-05-08
Prevent '#' character from being used in custom alias

### LINKED ISSUE #743: [BUG] Aliases with '#' character don't work
**Describe the bug**
If a short alias contains '#' character, then everything after the character is discarded when short link request is sent to server.

If short alias <word 1>#<word 2> is requested, then long link for <word 1> will be returned (if it exists) with "#<word 2>" appended to long link URL. If no long link for <word 1> exists then 404 will be returned.

**To Reproduce**
1. Create a short link with alias "alias#fragment" and point it to "http://localhost:2000"
2. Attempt to open short link, you'll get 404 (with "#fragment" appended to the end of 404 url)
3. Create another short link with alias "alias" and point it to "http://localhost:9000"
4. Attempt to open this short link, it will point to "http://localhost:9000" as expected
5. Attempt to open "alias#fragment" again, it will point to "http://localhost:9000/#fragment"

**Expected behavior**
'#' shouldn't be allowed in a short link alias.

**Screenshots**
(using different example alias)
Screen Shot 2020-05-06 at 4 56 57 PM

Screen Shot 2020-05-06 at 4 57 27 PM

result

**Additional context**
I believe this is because the fragment part is removed from URL before it's sent to server.

However, when short link is created, the alias is saved in full to DB. (ie. short link would be saved as <word 1>#<word 2> to DB) The problem pertains to the fact that the server when trying to transform aliased link to long link can only search for the <word 1> part of the link, the rest is discarded by the browser before the request hits the server.

Coteh:
Fixes #743.

### Description
Aliases created with URL fragment character '#' cannot be looked up since browser cuts off fragment from URL before it reaches server.

### Screenshots
See #743.

### Description
Aliases with '#' character are no longer allowed to be created. Additionally, a new frontend error for invalid alias will be shown when this happens.

### Screenshots
image


backend/app/usecase/url/urlcreator_test.go
```diff
@@ -56,13 +57,8 @@ func TestURLCreatorPersist_CreateURL(t *testing.T) {
 			expHasErr: true,
 		},
 		{
-			name: "alias too long",
-			urls: urlMap{
-				"220uFicCJj": entity.URL{
```
Coteh, line 61: @byliuyang Removed this URL entry from urlMap since it's not used for the test.

backend/app/usecase/url/urlcreator_test.go
```diff
@@ -134,13 +142,8 @@ func TestURLCreatorPersist_CreateURL(t *testing.T) {
 			},
 		},
 		{
-			name: "no available key",
-			urls: urlMap{
-				"220uFicCJj": entity.URL{
```
Coteh, line 139: @byliuyang Removed this URL entry from urlMap since it's not used for the test.

backend/app/usecase/url/urlcreator_test.go
```diff
@@ -24,6 +24,7 @@ func TestURLCreatorPersist_CreateURL(t *testing.T) {
 
 	alias := "220uFicCJj"
 	longAlias := "an-alias-cannot-be-used-to-specify-default-arguments"
+	invalidFragmentAlias := "cant-have#chr"
```
Coteh, line 27: @byliuyang As discussed in #741, we will need to update `CreateURL` to accept a string literal instead of string pointer if we want to hardcode aliases into the test cases directly

Coteh:
Integration test failure appears to be unrelated:
```
panic: Unable to create migration plan because of 15_create_user_role_table.sql: unknown migration in database [recovered]
	panic: Unable to create migration plan because of 15_create_user_role_table.sql: unknown migration in database

goroutine 9 [running]:
testing.tRunner.func1(0xc0000ee600)
	/usr/local/go/src/testing/testing.go:874 +0x3a3
panic(0xa498e0, 0xc00038de00)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/short-d/app/fw/db/dbtest.AccessTestDB(0xc11ce0, 0x10983a8, 0xc17900, 0x10983a8, 0xb2e4b5, 0xb, 0xc000026088, 0xe, 0x1ce8, 0xc000026048, ...)
	/go/pkg/mod/github.com/short-d/app@v0.0.0-20200505070402-e85f4ca85824/fw/db/dbtest/access.go:36 +0x2ac
github.com/short-d/short/backend/app/adapter/sqldb_test.TestChangeLogSql_GetChangeLog.func1(0xc0000ee600)
	/app/app/adapter/sqldb/changelog_integration_test.go:79 +0x106
testing.tRunner(0xc0000ee600, 0xc000157210)
	/usr/local/go/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:960 +0x350
FAIL	github.com/short-d/short/backend/app/adapter/sqldb	0.131s
```

byliuyang:
https://github.com/short-d/short/issues/546

byliuyang:
I think we should include input validation in the frontend with error message. For backend, we should return concrete error so that users know they can't use `#` tag.

Coteh:
> I think we should include input validation in the frontend with error message. For backend, we should return concrete error so that users know they can't use `#` tag.

So if I understand this correctly, we will have input validation for `#` tag in both the frontend and backend. Is this correct?

byliuyang:
@Coteh Yes! Can return meaningful errors so that the user know the problem is caused by including `#`.

Coteh:
Originally wanted to extend invalid alias error in backend to contain a slice array of invalid tokens, but the frontend error service will need a bit of redesign to handle extra information such as this. I decided to go with just a concrete error for fragment in backend instead because that's the only invalid character at the moment, and also added a check for fragment character in frontend too.

Ready for review.

Coteh:
@byliuyang Ready for review

backend/app/usecase/validator/alias.go
```diff
@@ -13,20 +16,21 @@ type CustomAlias struct {
 
 // IsValid checks whether the given alias has valid format.
 func (c CustomAlias) IsValid(alias *string) bool {
-	if alias == nil {
-		return true
-	}
-
-	if *alias == "" {
+	if alias == nil || *alias == "" {
```
byliuyang, line 19: Prefer separate if & else for simplicity

backend/app/usecase/url/urlcreator.go
```diff
@@ -73,6 +80,9 @@ func (c CreatorPersist) CreateURL(url entity.ShortLink, customAlias *string, use
 	}
 
 	if !c.aliasValidator.IsValid(customAlias) {
+		if c.aliasValidator.HasFragmentCharacter(*customAlias) {
+			return entity.ShortLink{}, ErrAliasWithFragment(*customAlias)
+		}
```
byliuyang, line 85: I think we should modify `IsValid` interface to return `boolean, Violation`. Violation is an enum. We are not 100% sure it's fragment violation just by calling this function here.

byliuyang: Then we can modify `ErrInvalidCustomAlias` include a field call `violation`

byliuyang: `ErrInvalidCustomAlias` has to be changed to a struct instead.

Coteh: Done

backend/app/usecase/validator/alias.go
Coteh: Done

Coteh:
Here's how the responses look for some of the violation errors: (including fragment character error)
Screen Shot 2020-05-22 at 8 03 09 PM
Screen Shot 2020-05-22 at 8 03 22 PM
Screen Shot 2020-05-22 at 8 03 36 PM

Also created an explicit `IsValid` interface in `validator.go`.

There is no error message emitted from backend to frontend. Atm, the frontend validation checks will catch the long alias and fragment character errors and display appropriate message so there is no need to do this from the backend IMO.

byliuyang:
@Coteh We cannot assume attackers are not injecting Javascript into our frontend in order to hack our server: https://martinfowler.com/articles/web-security-basics.html#RejectUnexpectedFormInput

backend/app/usecase/validator/violation.go
```diff
@@ -0,0 +1,25 @@
...
+func (v Violation) String() string {
+	return [...]string{
+		"Valid",
+		"InvalidAlias",
+		"InvalidLongLink",
+		"AliasTooLong",
+		"LongLinkTooLong",
+		"HasFragmentCharacter",
+	}[v]
+}
```
byliuyang, line 25: You can also change line 4 to `type Violation string`

backend/app/usecase/validator/alias_test.go
```diff
@@ -42,3 +42,37 @@ func TestCustomAlias_IsValid(t *testing.T) {
...
+			expIsValid: false,
+		},
+	}
+
+	validator := NewCustomAlias()
+	for _, testCase := range testCases {
+		testCase := testCase
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, testCase.expIsValid, validator.HasFragmentCharacter(testCase.alias))
```
byliuyang, line 75: Now we can change this to `isValid` and check the violation type!

backend/app/usecase/validator/alias.go
```diff
@@ -12,19 +15,29 @@ type CustomAlias struct {
...
+	if c.HasFragmentCharacter(*alias) {
+		return false, HasFragmentCharacter
 	}
-	return true
+
+	return true, Valid
+}
+
+// HasFragmentCharacter returns whether the alias contains the '#' character which starts fragment identifiers in URLs
+func (c CustomAlias) HasFragmentCharacter(alias string) bool {
```
byliuyang, line 39: I think we can rename it to `hasFragmentCharacter` because we will be using this function only inside `CustomAlias` struct.

backend/app/adapter/gqlapi/resolver/authmutation.go
```diff
@@ -67,9 +67,9 @@ func (a AuthMutation) CreateURL(args *CreateURLArgs) (*URL, error) {
 	case url.ErrAliasExist:
 		return nil, ErrURLAliasExist(*customAlias)
 	case url.ErrInvalidLongLink:
-		return nil, ErrInvalidLongLink(u.LongLink)
+		return nil, ErrInvalidLongLink{u.LongLink, err.(url.ErrInvalidLongLink).Violation.String()}
```
byliuyang, line 70:
```suggestion
		return nil, ErrInvalidLongLink{u.LongLink, string(err.(url.ErrInvalidLongLink).Violation)}
```

backend/app/adapter/gqlapi/resolver/authmutation.go
```diff
@@ -67,9 +67,9 @@ func (a AuthMutation) CreateURL(args *CreateURLArgs) (*URL, error) {
...
 	case url.ErrInvalidCustomAlias:
-		return nil, ErrInvalidCustomAlias(*customAlias)
+		return nil, ErrInvalidCustomAlias{*customAlias, err.(url.ErrInvalidCustomAlias).Violation.String()}
```
byliuyang, line 72: It was awhile back when I created Short. Now, in Go 13, `Is` and `As` are added to help check error. Please add a TODO to help migrate to the new syntax: https://blog.golang.org/go1.13-errors

backend/app/usecase/validator/alias_test.go
Coteh: TODO
- [ ] Remove `HasFragmentCharacter` test and add test for `isValid` checking for violation type

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: I think this suggestion will work once I change the type of `Violation` from `int` to `string` like you suggested above.

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: TODO
- [ ] Add TODO to migrate switch statement for error type checking to user `errors.As` instead

Coteh:
> @Coteh We cannot assume attackers are not injecting Javascript into our frontend in order to hack our server: https://martinfowler.com/articles/web-security-basics.html#RejectUnexpectedFormInput

I don't believe this PR introduces any attack vectors from frontend. The backend will simply check for the presence of a '#' character in the alias and send response to server with an error message with a 'HasFragmentCharacter' violation.

All "invalid alias" messages from backend will look like this when presented on frontend:

Screen Shot 2020-05-22 at 7 59 31 PM

Is this sufficient? Or should we add some extra information about the `violation`? (`violation` being a fixed value determined by backend and not controllable by attacker)

byliuyang:
@Coteh I think we should mention the violation to avoid confusions.

byliuyang:
@Coteh Actually, I am wondering whether we should reveal violation the frontend directly. As you mentioned, we also have validation here. Frontend is for clarity, backend security. What do you think?

Coteh:
I think what we have right now in this PR is good. There are frontend validation checks, and there are also backend validation checks in the case that we call the API directly, or in the case of a future client like a mobile app that doesn't have any frontend validation checks implemented (although it can have them implemented).

The `violation` field that is emitted from the response is sufficient to tell the user what kind of violation error occurred, and it is not too revealing either. (the violation string is essentially hard-coded response to a certain condition) The field is also flexible - we don't necessarily need to emit it for every type of validation error there is, which is important to not let attackers know that their attack attempts have been acknowledged.

byliuyang:
@Coteh Aha, for TODO I mean adding this in the code so that it reminds people to address tech debts when reading the code: TODO format: // TODO(issue#631): support responsive layouts.

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -77,16 +78,20 @@ func (e ErrNotHuman) Error() string {
...
 var _ GraphQlError = (*ErrInvalidLongLink)(nil)
 
 // Extensions keeps structured error metadata so that the clients can gracefully
 // handle the error.
 func (e ErrInvalidLongLink) Extensions() map[string]interface{} {
 	return map[string]interface{}{
-		"code":     ErrCodeInvalidLongLink,
-		"longLink": string(e),
+		"code":      ErrCodeInvalidLongLink,
+		"longLink":  string(e.longLink),
```
byliuyang, line 93:
```suggestion
		"longLink":  e.longLink,
```

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -77,16 +78,20 @@ func (e ErrNotHuman) Error() string {
...
+		"violation": string(e.violation),
```
byliuyang, line 94:
```suggestion
		"violation": e.violation,
```

backend/app/adapter/gqlapi/resolver/error.go
byliuyang: I think they are already strings based on line 82 to 83.

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -106,7 +114,8 @@ var _ GraphQlError = (*ErrInvalidCustomAlias)(nil)
 func (e ErrInvalidCustomAlias) Extensions() map[string]interface{} {
 	return map[string]interface{}{
 		"code":        ErrCodeInvalidCustomAlias,
-		"customAlias": string(e),
+		"customAlias": string(e.customAlias),
```
byliuyang, line 117:
```suggestion
		"customAlias": e.customAlias,
```

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -106,7 +114,8 @@ var _ GraphQlError = (*ErrInvalidCustomAlias)(nil)
...
+		"violation":   string(e.violation),
```
byliuyang, line 118:
```suggestion
		"violation":   e.violation,
```

backend/app/usecase/validator/alias_test.go
byliuyang: @Coteh Please note that `HasFragmentCharacter` (public) should be renamed to `hasFragmentCharacter ` (private)

backend/app/usecase/validator/violation.go
```diff
@@ -0,0 +1,25 @@
+package validator
+
+// Violation represents a type of invalid error encountered by the validator.
+type Violation int
+
+const (
+	Valid Violation = iota
+	InvalidAlias
+	InvalidLongLink
```
byliuyang, line 9: We should make those violation provide extract context to the API consumers. Not sure how will they react when seeing `InvalidAlias`

byliuyang:
@Coteh Then we would just return invalid long link error from the API but keep the violation error in the backend so that we can logged them for security analysis when issues come up.

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: My bad, I forgot to address this. TODO has been added and my next commit will contain the TODO comments.

backend/app/adapter/gqlapi/resolver/error.go
Coteh: Done

backend/app/adapter/gqlapi/resolver/error.go
Coteh: Done

backend/app/adapter/gqlapi/resolver/error.go
Coteh: Done

backend/app/adapter/gqlapi/resolver/error.go
Coteh: Done

backend/app/usecase/validator/violation.go
Coteh: What kind of context are you expecting? Should they see the custom alias they tried to enter?

byliuyang: @Coteh I am not sure whether invalid long link a violation. Long link too long, long link empty are both violations. Invalid long link didn’t seem to communicate anything. The API consumer would have no clue how to handle it. Maybe we can consider removing violation code ( keep validation code but just not returning what is violating ) in the backend. What do you think?

byliuyang: Are we able to replace those 2 lines with concrete errors?

Coteh: @byliuyang Sorry for late response. My original intention with `InvalidLongLink` and `InvalidAlias` were to be general catch-all error types for invalid errors that didn't have a concrete type defined yet. I understand your point and I think we can probably just return no "violation" error instead for these cases like you suggest.

backend/app/usecase/validator/violation.go
Coteh: Forgot to address this. 🙈 Thanks for reminding me!

backend/app/usecase/validator/alias.go
Coteh: Done

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: Done

backend/app/usecase/validator/violation.go
Coteh: @byliuyang In the case that the long link or custom alias validation error contains no corresponding violation code, should the `extensions` field from the response contain no `violation` field? Would this be better than having `InvalidLongLink` and `InvalidAlias` as catch-all violation types? The benefit of omitting the `violation` field in this case is that the API consumer will only need to check whether the field is present in the response instead of checking for the catch-all types specifically. The downside is that it won't be 100% certain whether the `violation` field will be present in the error response.

byliuyang: @Coteh How about this. Our use case will return the exact violation error so that we can track those metrics in DataDog. However, we API never return violation.

byliuyang: Or we just complete remove the violation reason from the backend because frontend can already provide the user some hints on what's going wrong.

PR #757, opened 2020-05-10
Add Authorizer to manager user roles and permission

### LINKED ISSUE #729: [Feature] Manage access control through authorizer

byliuyang:
Fixes #729


backend/app/adapter/sqldb/userrole_integration_test.go
```diff
@@ -0,0 +1,227 @@
...
+			name: "add 1 role for nonexistent user",
+			user: entity.User{
+				ID: "1343",
+			},
+			userRoleRows: []userRoleTableRow{
+				{"4444", role.Basic},
+			},
+			newRoles:      []role.Role{role.ChangeLogViewer},
+			expectedRoles: []role.Role{role.ChangeLogViewer},
+			hasErr:        false,
```
arberiii, line 50: shouldn't hasErr be true?

byliuyang: We didn't have foreign key constraint yet. Will fix after we migrated from email to user ID.

Issue #643, opened 2020-04-08
[Feature] Integrate change log api with the frontend


byliuyang:
**What is the problem?**
[Change log service](https://github.com/short-d/short/blob/master/frontend/src/service/ChangeLog.service.ts) in the frontend is rendering dummy data hard-coded. 

**Your solution**
Integrate change log service with [GraphQL API](https://github.com/short-d/short/blob/033ccc686046e12f72cdfdbdd3b829bebfab88e1/backend/app/adapter/graphql/schema.go#L17) so that change log modal works as a whole.

kapillamba4:
email must be mappped to user id before this issue

byliuyang:
@kapillamba4 How about we use email as user ID for now?

byliuyang:
The ID migration is taking more time than I expected.

byliuyang:
We pass in `entity.User` to the use case instead of `email` itself to easy changes.

byliuyang:
Done.

PR #766, opened 2020-05-12
Implement ChangeLogGraphQLApi

kapillamba4:
Part of #643


frontend/src/component/pages/HomePage.tsx
```diff
@@ -107,16 +108,20 @@ export class HomePage extends Component<Props, State> {
...
-      }
-
-      this.showChangeLogs();
-    });
+        this.showChangeLogs();
+      });
+    } catch (err) {
+      const { changeLogErr } = err;
+      this.props.store.dispatch(raiseGetChangeLogError(changeLogErr));
+    }
```
byliuyang, line 124: Do we want to load change log when component is loading or wait until user tries to open the modal? I recommend we wrap this into separate function: `autoShowChangeLog`, `autoShowChangeLog`, or something else to reduce complexity.

frontend/src/component/ui/ChangeLogModal.tsx
```diff
@@ -44,10 +44,10 @@ export class ChangeLogModal extends Component<Props, State> {
     return (
       <div className={'changelog'}>
         <ul>
-          {changeLog.map((update: Update) => (
+          {changeLog.map((update: Change) => (
```
byliuyang, line 47:
```suggestion
          {changeLog.map((change: Change) => (
```

frontend/src/dep.ts
```diff
@@ -59,7 +60,15 @@ export function initUIFactory(
   const versionService = new VersionService(envService);
   const store = initStore();
   const searchService = new SearchService();
-  const changeLogService = new ChangeLogService();
+  const changeLogGraphQLApiService = new ChangeLogGraphQLApi(
+    authService,
+    envService,
+    graphQLService
+  );
```
byliuyang, line 67: I saw you are creating a separate GraphQL API service for change log. Is this to prepare for spanning it to a new service? We have a generic GraphQL API client here: https://github.com/short-d/short/blob/master/frontend/src/service/ShortGraphQL.api.ts

frontend/src/entity/Change.tsx
```diff
@@ -0,0 +1,6 @@
+export interface Change {
+  id: string;
+  title: string;
+  summaryMarkdown: string;
```
byliuyang, line 4: I think this field is optional: https://github.com/short-d/short/blob/master/backend/app/adapter/gqlapi/schema.go#L31

frontend/src/entity/ChangeLog.tsx
```diff
@@ -0,0 +1,6 @@
+import { Change } from './Change';
+
+export interface ChangeLog {
+  changes: Change[];
+  lastViewedAt: string;
```
byliuyang, line 5: I think this is also optional: https://github.com/short-d/short/blob/master/backend/app/adapter/gqlapi/schema.go#L25

frontend/src/service/ChangeLog.service.ts
```diff
@@ -1,73 +1,66 @@
...
-        return;
-      }
-
-      if (!changeLog[0]) {
-        resolve(false);
-        return;
-      }
+      try {
+        const changeLog = await this.changeLogGraphQLApi.getChangeLog();
+        if (!changeLog.changes) {
```
byliuyang, line 15:
```suggestion
        if (!changeLog.changes || changeLog.changes.length < 1) {
```

frontend/src/service/ChangeLog.service.ts
```diff
@@ -1,73 +1,66 @@
...
+        }
 
-      if (lastSeenAt >= changeLog[0].releasedAt) {
-        resolve(false);
-        return;
-      }
+        if (!changeLog.changes[0]) {
+          resolve(false);
+          return;
+        }
```
byliuyang, line 23:
```suggestion
```

frontend/src/service/ChangeLog.service.ts
```diff
@@ -1,73 +1,66 @@
...
-  }
+        for (let i = 0; i < changeLog.changes.length; i++) {
+          if (
+            new Date(changeLog.lastViewedAt).getTime() <
+            new Date(changeLog.changes[i].releasedAt).getTime()
+          ) {
+            resolve(true);
+            return;
+          }
+        }
```
byliuyang, line 38: You can do a simple [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) and find the most recent releaseAt. Then compare it with `lastViewedAt`. Separate the logics not only improves readability but also allow us to optimize the algorithm in the future.

byliuyang: We will have to sort the changes based on timestamp when displaying them eventually.

byliuyang: We can create a helper method in this class to fetch the change log and sort them.

frontend/src/service/ChangeLogGraphQL.api.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private gqlErrorToCode(graphQLError: IGraphQLError): string {
+    switch (graphQLError.extensions.code) {
+      case Err.Unauthenticated:
+        return Err.Unauthenticated;
+      default:
+        return Err.Unknown;
+    }
+  }
```
byliuyang, line 112: nit: We may want to reuse this across GraphQL APIs

frontend/src/dep.ts
kapillamba4: Resolved on slack, https://github.com/short-d/short/issues/770

frontend/src/entity/Change.tsx
```diff
@@ -0,0 +1,6 @@
+export interface Change {
+  id: string;
+  title: string;
+  summaryMarkdown?: string;
+  releasedAt: string;
```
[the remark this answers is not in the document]
byliuyang: I recommend converting the string to date inside changelog API so that we don't have to parse them again for sorting.

frontend/src/service/ChangeLogGraphQL.api.ts
```diff
@@ -0,0 +1,89 @@
...
+          const errCodes = getErrorCodes(err);
+          reject(errCodes[0]);
+        });
+    });
+  }
+
+  private parseChangeLog(changeLog: IGraphQLChangeLog): ChangeLog {
+    return {
+      changes: changeLog.changes.map(this.parseChange),
+      lastViewedAt: changeLog.lastViewedAt
```
byliuyang, line 77: Convert to date when not `null`.

frontend/src/service/ChangeLogGraphQL.api.ts
```diff
@@ -0,0 +1,89 @@
...
+    };
+  }
+
+  private parseChange(change: IGraphQLChange): Change {
+    return {
+      id: change.id,
+      title: change.title,
+      summaryMarkdown: change.summaryMarkdown,
+      releasedAt: change.releasedAt
```
byliuyang, line 86: Convert to date when not `null`.

frontend/src/component/ui/ChangeLogModal.spec.tsx
```diff
@@ -15,19 +15,25 @@ describe('ChangeLogModal', () => {
     const changeLogModalRef = React.createRef<ChangeLogModal>();
     const changeLog = [
       {
+        id: '12345',
         title: 'Lorem ipsum',
-        releasedAt: 1500000000003,
-        summary: 'Lorem ipsum dolor sit amet, consectetuer adipiscing elit'
+        releasedAt: new Date(1500000000003),
```
byliuyang, line 20: I recommend doing https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse so that the dates are readable.

frontend/src/service/ChangeLog.service.ts
```diff
@@ -1,73 +1,72 @@
...
   hasUpdates(): Promise<boolean> {
     return new Promise(async (resolve, reject) => {
-      const changeLog = await this.getChangeLog();
-      const lastSeenAt = await this.getLastSeenChangeLog();
-      if (!changeLog) {
-        resolve(false);
-        return;
+      let changeLog;
+      try {
+        changeLog = await this.changeLogGraphQLApi.getChangeLog();
```
byliuyang, line 16: I though if you move the change attribute validation into `getChangeLog` and call `this.getChangeLog()` here, then you don't have to sort again in this function. Please let me know if I am wrong.

frontend/src/service/ChangeLogGraphQL.api.ts
```diff
@@ -0,0 +1,95 @@
...
+      changes: changeLog.changes.map(this.parseChange)
+    };
+  }
+
+  private parseChange(change: IGraphQLChange): Change {
+    return {
+      id: change.id,
+      title: change.title,
+      summaryMarkdown: change.summaryMarkdown,
+      releasedAt: new Date(change.releasedAt)
```
byliuyang, line 92: What happens if `change.releasedAt` is actually null?

kapillamba4: I don't think change.releasedAt can be null according to graphQL schema

byliuyang: Okey.

frontend/src/service/ChangeLog.service.ts
byliuyang: Discussed offline. Keep it as it is.

PR #773, opened 2020-05-13
Parallelize automated tests

byliuyang:


backend/app/adapter/gqlapi/resolver/authquery_test.go
```diff
@@ -85,7 +86,9 @@ func TestAuthQuery_URL(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
+		testCase := testCase
```
arberiii, line 89: why is this needed?

PR #786, opened 2020-05-15
Use UTC time when creating new Change

kapillamba4:
Server time may not be in UTC and therefore when new change is created released_at may not be in UTC as well. In order to maintain same time irrespective of server location we should use enforce UTC time


byliuyang:
This is to ensure we have consistent timing across services.

PR #788, opened 2020-05-15
Add ID migration tool

byliuyang:
### Description
Previously we are migrating from email to user ID. However, after the first and the second DB migrations, we unfortunately lost the most recent user data and had to redo the migration.

### Description
Add ID migration tool so that we can re-perform the migration.

### DIFF
--- a/backend/cmd/cmd.go
+++ b/backend/cmd/cmd.go
@@ -25,8 +25,7 @@ func NewRootCmd(
 		cli.CommandConfig{
 			Usage:        "start",
 			ShortHelpMsg: "Start service",
-			// TODO(issue#789): remove the star from cmd signature
-			OnExecute: func(cmd *cli.Command, args []string) {
+			OnExecute: func(cmd cli.Command, args []string) {
 				config.MigrationRoot = migrationRoot
 				app.Start(
 					dbConfig,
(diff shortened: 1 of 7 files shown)


backend/cmd/cmd.go
```diff
@@ -41,6 +43,23 @@ func NewRootCmd(
 		"migration migrations root directory",
 	)
 
+	idCmd := cmdFactory.NewCommand(cli.CommandConfig{
+		Usage:        "data-id",
+		ShortHelpMsg: "Use user ID to uniquely identify a user",
+		OnExecute: func(cmd *cli.Command, args []string) {
```
arberiii, line 49: cli.Command is interface therefore a pointer itself

backend/cmd/cmd.go
```diff
@@ -41,6 +43,23 @@ func NewRootCmd(
...
+			kgsConfig := provider.KgsRPCConfig{
+				Hostname: config.KgsHostname,
+				Port:     config.KgsPort,
+			}
+			keyGenBufferSize := provider.KeyGenBufferSize(config.KeyGenBufferSize)
+			dataTool, err := dep.InjectDataTool(dbConfig, dbConnector, keyGenBufferSize, kgsConfig)
+			if err != nil {
+				panic(err)
```
arberiii, line 57:
```suggestion
fmt.Println(err)
os.Exit(1)
```

byliuyang: @arberiii Why not panic? I think it's okey to crash if we running the tool offline.

backend/cmd/cmd.go
byliuyang: @arberiii Yes?

backend/cmd/cmd.go
arberiii: I thought to keep the same error handling through out the function

backend/cmd/cmd.go
arberiii: I thought we don’t to pass the pointer, like just cli.Command. What do you think?

backend/cmd/cmd.go
byliuyang: Done.

backend/cmd/cmd.go
byliuyang: @arberiii Adde TODO to update the signature in the app framework.

backend/tool/data.go
```diff
@@ -0,0 +1,58 @@
...
+// Data transform existing data to the new format and move them to the correct
+// location.
+type Data struct {
+	keyGen keygen.KeyGenerator
+	db     *sql.DB
+}
+
+// EmailToID generates IDs for users and changes DB tables reference those IDs.
+func (d Data) EmailToID() {
+	rows, err := d.db.Query(`
```
alldroll, line 19: I suggest using batch reading

```
SELECT email FROM "user" where "id is empty" LIMIT 1000
```

Here "id is empty" means, that we fetch only records with unsettled `id`. We perform the above query again and again until after the select dataset size is less or equal to 1000.

It gives us a feeling to be sure that we can handle this migration even for a big number of records "just a good practice"

backend/cmd/cmd.go
alldroll: I think we should `fmt.Error` with a `%w` verb.

backend/tool/data.go
byliuyang: @alldroll Should we automatically increase the number of counts using for loops or we give admin the control to decide the size?

alldroll: It depends but in general, we increase the number of counts. I think the simplest way is the best one:)

backend/cmd/cmd.go
byliuyang: @alldroll Wa are directly printing the error here. 'fmt.Errorf' returns unused result and didn't provide much value in this case. I think it's better to use `fmt.Errorf` inside the application logic.

backend/tool/data.go
byliuyang: @alldroll Done.

Issue #753, opened 2020-05-08
[BUG] Potentially broken test in urlcreator_test


Coteh:
Wasn't sure whether to file this as "Refactor" or "BUG", feel free to refile it if needed. Or close it if there's a misunderstanding. Thanks.

**Describe the bug**
"alias exists" test in `urlcreator_test.go` still passes even after the URL entry in test case's `urlMap` is removed.

**To Reproduce**
1. In "alias exists" test case in `urlcreator_test.go`, replace:
```
urls: urlMap{
	"220uFicCJj": entity.URL{
		Alias:    "220uFicCJj",
		ExpireAt: &now,
	},
},
```
with
```
urls: urlMap{}
``` 

**Expected behavior**
The test should pass because the URL is defined within `urlMap` of test case.

**Additional context**
The `alias exists` error's `Error()` string is also an empty string, giving no indication of what kind of error happened. See below:

```
=== RUN   TestURLCreatorPersist_CreateURL
=== PAUSE TestURLCreatorPersist_CreateURL
=== CONT  TestURLCreatorPersist_CreateURL
=== RUN   TestURLCreatorPersist_CreateURL/alias_exists
=== PAUSE TestURLCreatorPersist_CreateURL/alias_exists
=== RUN   TestURLCreatorPersist_CreateURL/alias_too_long
=== PAUSE TestURLCreatorPersist_CreateURL/alias_too_long
=== RUN   TestURLCreatorPersist_CreateURL/create_alias_successfully
=== PAUSE TestURLCreatorPersist_CreateURL/create_alias_successfully
=== RUN   TestURLCreatorPersist_CreateURL/automatically_generate_alias_if_null_alias_provided
=== PAUSE TestURLCreatorPersist_CreateURL/automatically_generate_alias_if_null_alias_provided
=== RUN   TestURLCreatorPersist_CreateURL/automatically_generate_alias_if_empty_string_alias_provided
=== PAUSE TestURLCreatorPersist_CreateURL/automatically_generate_alias_if_empty_string_alias_provided
=== RUN   TestURLCreatorPersist_CreateURL/no_available_key
=== PAUSE TestURLCreatorPersist_CreateURL/no_available_key
=== CONT  TestURLCreatorPersist_CreateURL/alias_exists
=== CONT  TestURLCreatorPersist_CreateURL/automatically_generate_alias_if_null_alias_provided
    TestURLCreatorPersist_CreateURL/alias_exists: urlcreator_test.go:191: err-> 
[...]

PR #796, opened 2020-05-20
Broken test in urlcreator_test

arberiii:
Fixes [#753](https://github.com/short-d/short/issues/753)


Coteh:
Test failing due to #546, not related to this change.

arberiii:
> Test failing due to #546, not related to this change.

If you follow the lifecycle of that particular test case, it tries to create a URL using the function creator.CreateURL with first parameter testCase.url, which is empty, and then it fails the condition that long link is non-empty and returns the ErrInvalidLongLink(longLink) which also creates an error with an empty message. Later on, that function is checked whether the alias exists.

backend/app/usecase/validator/longlink.go
```diff
@@ -19,6 +19,11 @@ func (l LongLink) IsValid(longLink *string) bool {
 		return false
 	}
 
+	return true
```
byliuyang, line 22: Nit: Directly return `*longLink != ""`

backend/app/usecase/url/urlcreator_test.go
```diff
@@ -50,7 +50,9 @@ func TestURLCreatorPersist_CreateURL(t *testing.T) {
 			user: entity.User{
 				Email: "alpha@example.com",
 			},
-			url:       entity.URL{},
+			url: entity.URL{
+				OriginalURL: "https://www.google.com",
+			},
 			isPublic:  false,
 			expHasErr: true,
```
byliuyang, line 57: We will have to check error types in the future. Add a TODO sounds good to me.

backend/app/usecase/shortlink/creator_test.go
```diff
@@ -51,7 +52,9 @@ func TestShortLinkCreatorPersist_CreateShortLink(t *testing.T) {
 			user: entity.User{
 				Email: "alpha@example.com",
 			},
-			shortLink: entity.ShortLink{},
+			shortLink: entity.ShortLink{
```
arberiii, line 55: If the long link is not provided, this test case will have an error because the long link is not provided. If any long link is provided then the test case has an error because the alias exist as intended!

PR #799, opened 2020-05-21
Add type column to feature toggle table

rohithbalaji123:
Part of #612 

### Description
Add a column "type" in feature toggle table to store the different types of features.
<del>Currently, NULL value in the column means default feature toggle and "permission" value in column means the feature depends on the role of the user requesting the feature. When supporting admin panel, if we are allowing the user to update the type, I guess it will be better to replace the NULL into some string which refers the toggle to a feature toggle.</del>

"manual" value for type column means the default feature toggle and "permission" value for type column means the feature toggle depends on the role of the user requesting the feature.


byliuyang:
I think we should not allow NULL. Default toggle will be "manual" toggle.

backend/app/entity/toggle.go
```diff
@@ -1,7 +1,14 @@
 package entity
 
+import "database/sql"
```
byliuyang, line 3: No SQL in either entity or use case. They are core business logics. Should not depends on specific technology. Let's use enum instead: https://blog.learngoprogramming.com/golang-const-type-enums-iota-bc4befd096d3

PR #798, opened 2020-05-21
Add updateURL mutation to schema

oatovar:
* Add `updateURL` endpoint to graphql api schema and implement resolver.

* Fix bug in url update statement that causes update to fail persistance.

* Update wired dependencies.

PR #807, opened 2020-05-23
Change internal CreateURL interfaces to accept string rather than *string

Coteh:
Currently, you are allowed to pass no value for `customAlias` to the GraphQL API and it'll be treated as `null` and a random alias is generated. However, since empty string for custom alias also denotes random alias, there is no reason to have the `customAlias` parameter be nullable.

This will also make it easier for various interfaces and tests to be understood.


byliuyang:
Does this mean we are forcing API consumer to provide empty alias?

Coteh:
Yes. If this is not desirable, I can close.

On May 22, 2020, at 8:49 PM, Harry Liu <notifications@github.com<mailto:notifications@github.com>> wrote:


Does this mean we are forcing API consumer to provide empty alias?

byliuyang:
@Coteh I think we should have a discussion. I like the idea but not sure whether this will cause inconvenience for API consumer. We may expose APIs to the external app in the future.

Coteh:
I think you’re right. It would be counterintuitive to have to specify “” for `customAlias` if a custom alias is not desired. I think I’ll rework this PR so that custom alias is still a nullable field but there will be check for null in the handler that will turn it into empty string before entering `CreateURL` method.

On May 22, 2020, at 8:59 PM, Harry Liu <notifications@github.com<mailto:notifications@github.com>> wrote:


@Coteh<https://github.com/Coteh> I think we should have a discussion. I like the idea but not sure whether this will cause inconvenience for API consumer. We may expose APIs to the external app in the future.

Coteh:
I gave it some thought, and I think this is a nice alternative. The GraphQL API will still accept `null` value for `customAlias` which is expected when customer of API does not want a custom alias, and before `CreateURL` is called on the backend, the mutation handler will convert `nil` custom aliases to empty string ("") aliases. This will allow tests to be more terse without too much modification of backend implementation, and without any modification of consumer-facing API.

backend/app/usecase/shortlink/creator_test.go
```diff
@@ -46,7 +44,7 @@ func TestShortLinkCreatorPersist_CreateShortLink(t *testing.T) {
 					ExpireAt: &now,
 				},
 			},
-			alias: &alias,
+			alias: alias,
```
arberiii, line 47:
```suggestion
			alias: "220uFicCJj",
```
We can put the alias value "220uFicCJj" here direct

backend/app/usecase/shortlink/creator_test.go
```diff
@@ -74,7 +72,7 @@ func TestShortLinkCreatorPersist_CreateShortLink(t *testing.T) {
 		{
 			name:       "create alias successfully",
 			shortLinks: shortLinks{},
-			alias:      &alias,
+			alias:      alias,
```
arberiii, line 75:
```suggestion
			alias:      "220uFicCJj",
```

backend/app/usecase/shortlink/creator_test.go
```diff
@@ -141,7 +119,7 @@ func TestShortLinkCreatorPersist_CreateShortLink(t *testing.T) {
 				},
 			},
 			availableKeys: []keygen.Key{},
-			alias:         nil,
+			alias:         "",
```
arberiii, line 122:
```suggestion

```

backend/app/usecase/validator/longlink.go
```diff
@@ -10,20 +10,16 @@ type LongLink struct {
 }
 
 // IsValid checks whether the given long link has valid format.
-func (l LongLink) IsValid(longLink *string) bool {
-	if longLink == nil {
+func (l LongLink) IsValid(longLink string) bool {
+	if longLink == "" {
```
arberiii, line 14:
```suggestion
	if len(longLink) == 0 {
```
Do you think checking for length makes it easier to read and is the same as the check below for max length?

Coteh: I'm not sure if it makes it any easier to read, but I do agree that it makes it more consistent with the other length check. 👍

Coteh:
@byliuyang I agree. I was reading this PR https://github.com/short-d/short/pull/798 where you also mention this, and I think that with `OriginalURL` going to become a `*string` as well, we might be better off to close this PR and leave things the way they are for the time being. What do you think?

byliuyang:
@Coteh Let’s close this PR for now. I think in the future, we may be better off by introducing a separate input struct with all fields as optional.

PR #827, opened 2020-06-03
Make feature toggle decision at mount time instead of build time

rohithbalaji123:
Part of #612 

### Description
The feature toggle and auth toggle decision is currently being made in constructor of UIFactory making it happen only once during the initial setup unless the page is reloaded. With the current dynamic(auth based) feature toggle functionality in place, this won't work as users will keep alternating between logged in and out state. That is, when a user is initially logged out, the permission based features will be unavailable and it will continue to be unavailable even after logging in.

### Description
Instead of passing the decision(`Promise<bool>`) in the `withFeatureToggle` and `withPageAuth` HOCs, we pass the decision maker so that it can be called while mounting the component and thus satisfying the dynamicity.

### DIFF
--- a/frontend/src/component/hoc/withFeatureToggle.tsx
+++ b/frontend/src/component/hoc/withFeatureToggle.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 export default function(
   WrappedComponent: React.ComponentType<any>,
-  featureDecision: Promise<boolean>
+  featureDecision: () => Promise<boolean>
 ): React.ComponentType<any> {
   interface IState {
     isFeatureEnabled: boolean;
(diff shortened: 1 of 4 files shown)


frontend/src/component/hoc/withFeatureToggle.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 export default function(
   WrappedComponent: React.ComponentType<any>,
-  featureDecision: Promise<boolean>
+  featureDecision: () => Promise<boolean>
```
byliuyang, line 5:
```suggestion
  makeFeatureDecision: () => Promise<boolean>
```

Issue #809, opened 2020-05-23
[Feature] Implement gRPC API for Social Media Sharing Customization


kapillamba4:
**What is the problem?**
All the short links shared on social media platforms have the same summary card:
![Screenshot from 2020-05-20 21-38-56](https://user-images.githubusercontent.com/20877996/82735452-8bab9600-9d3f-11ea-89fa-60886f12eaae.png)

**Your solution**
Enable marketers to customize the summary card of the short links they shared on Facebook, Twitter, and LinkedIn.

**Additional context**
https://short-d.com/r/social-media-sharing

PR #830, opened 2020-06-03
Add sqldb adapters for updating meta tags

kapillamba4:
Part of #809


kapillamba4:
Splitting into 3 PR's

PR #835, opened 2020-06-04
Add delete option for change log repo

rohithbalaji123:
Part of #612 

### Description
Add delete APIs in ChangeLog repository and changeLog sqlDB adaptor.

### DIFF
--- a/backend/app/adapter/sqldb/changelog.go
+++ b/backend/app/adapter/sqldb/changelog.go
@@ -75,6 +75,20 @@ VALUES ($1, $2, $3, $4);
 	return newChange, nil
 }
 
+// DeleteChange removes a change based on a given ID
+func (c ChangeLogSQL) DeleteChange(ID string) error {
+	statement := fmt.Sprintf(`
+DELETE FROM "%s"
+WHERE "%s" = $1;
+`,
+		table.ChangeLog.TableName,
+		table.ChangeLog.ColumnID,
+	)
+
+	_, err := c.db.Exec(statement, ID)
+	return err
+}
+
 // NewChangeLogSQL creates ChangeLogSQL
 func NewChangeLogSQL(db *sql.DB) ChangeLogSQL {
 	return ChangeLogSQL{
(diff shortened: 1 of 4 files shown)


kapillamba4:
@rohithbalaji123 what will the graphQL schema look like for deleteChangeLog mutation ?

rohithbalaji123:
@kapillamba4 I was thinking of making it return the ID again if successful or `nil` when failed.
```
type AuthMutation {
	deleteChange(ID: String!): String
}
```

I was a bit inclined towards `ID/nil` instead of `true/false` because, it gives a bit more context while serving the same purpose. What do you think?

kapillamba4:
@rohithbalaji123, i agree returning `ID/nil` seems better option here

backend/app/adapter/sqldb/changelog.go
```diff
@@ -75,6 +75,20 @@ VALUES ($1, $2, $3, $4);
 	return newChange, nil
 }
 
+// DeleteChange removes a change based on a given ID
```
byliuyang, line 78:
```suggestion
// DeleteChange removes a change with the given ID
```

backend/app/usecase/repository/change_log_fake.go
```diff
@@ -30,6 +30,31 @@ func (c *ChangeLogFake) CreateChange(newChange entity.Change) (entity.Change, er
 	return newChange, nil
 }
 
+// DeleteChange removes a change based on a given ID
+func (c *ChangeLogFake) DeleteChange(ID string) error {
+	foundIdx := -1
+	for idx, change := range c.changeLog {
+		if change.ID == ID {
+			foundIdx = idx
```
byliuyang, line 38:
```suggestion
			return c.removeChangeAt(id)
```

backend/app/usecase/repository/change_log_fake.go
```diff
@@ -30,6 +30,31 @@ func (c *ChangeLogFake) CreateChange(newChange entity.Change) (entity.Change, er
...
+		}
+	}
+
+	if foundIdx == -1 {
+		return nil
+	}
+
+	return c.removeChangeAt(foundIdx)
```
byliuyang, line 46:
```suggestion
	return nil
```

byliuyang:
Hi @rohithbalaji123 , thank you so much for working on this feature! Left few comments on the PR.

PR #841, opened 2020-06-05
Add ByCreatedTimeASC order for Search

arberiii:
### Description
The result from the search doesn't have any order option other than Unchanged which is the default option.

### Description
Add ByCreatedTimeASC option, which sorts the result of the searchable resources based on created time in ascending order. 
Sorry for this rather larger PR, most of it is test cases. Besides the unit test for ByCreatedTimeASC order, in this PR I included the unit tests for Unchanged order which consists of the same test cases, but we can discuss if those tests are not meaningful for Unchanged order.

### DIFF
--- a/backend/app/usecase/search/order/created_time.go
+++ b/backend/app/usecase/search/order/created_time.go
@@ -13,6 +13,8 @@ type CreatedTime struct {
 }
 
 // ArrangeShortLinks arranges shortLinks based on CreatedAt.
+// The returned short links satisfies the created before comparision if CreatedAt
+// is not nil, otherwise it sits on the end of the slices if CreatedAt is nil.
 func (c CreatedTime) ArrangeShortLinks(shortLinks []entity.ShortLink) []entity.ShortLink {
 	sort.SliceStable(shortLinks, func(i, j int) bool {
 		if shortLinks[j].CreatedAt == nil {
@@ -26,6 +28,8 @@ func (c CreatedTime) ArrangeShortLinks(shortLinks []entity.ShortLink) []entity.S
 }
 
 // ArrangeUsers arranges users based on CreatedAt.
+// The returned users satisfies the created before comparision if CreatedAt
+// is not nil, otherwise it sits on the end of the slices if CreatedAt is nil.
 func (c CreatedTime) ArrangeUsers(users []entity.User) []entity.User {
 	sort.SliceStable(users, func(i, j int) bool {
 		if users[j].CreatedAt == nil {
--- a/backend/app/usecase/search/order/created_time.go
+++ b/backend/app/usecase/search/order/created_time.go
@@ -2,42 +2,45 @@ package order
 
 import (
 	"sort"
+	"time"
 
 	"github.com/short-d/short/backend/app/entity"
 )
 
 var _ Order = (*CreatedTime)(nil)
 
-// CreatedTime arranges searchable resources based on their created time.
+// CreatedTime arranges searchable resources based on when they are created.
 type CreatedTime struct {
 }
 
 // ArrangeShortLinks arranges shortLinks based on CreatedAt.
-// The returned short links satisfies the created before comparision if CreatedAt
-// is not nil, otherwise it sits on the end of the slices if CreatedAt is nil.
 func (c CreatedTime) ArrangeShortLinks(shortLinks []entity.ShortLink) []entity.ShortLink {
-	sort.SliceStable(shortLinks, func(i, j int) bool {
-		if shortLinks[j].CreatedAt == nil {
-			return true
-		} else if shortLinks[i].CreatedAt == nil {
-			return false
-		}
-		return shortLinks[i].CreatedAt.Before(*shortLinks[j].CreatedAt)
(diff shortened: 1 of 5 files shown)


byliuyang:
@arberiii I am fine with large unit test PRs! As long as the PR does contain too much complex business logic.

backend/app/usecase/search/order/created_time.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"sort"
+
+	"github.com/short-d/short/backend/app/entity"
+)
+
+var _ Order = (*CreatedTime)(nil)
+
+// CreatedTime arranges searchable resources based on their created time.
```
byliuyang, line 11:
```suggestion
// CreatedTime arranges searchable resources based on when they are created.
```

backend/app/usecase/search/order/created_time.go
```diff
@@ -0,0 +1,43 @@
...
+type CreatedTime struct {
+}
+
+// ArrangeShortLinks arranges shortLinks based on CreatedAt.
+// The returned short links satisfies the created before comparision if CreatedAt
+// is not nil, otherwise it sits on the end of the slices if CreatedAt is nil.
+func (c CreatedTime) ArrangeShortLinks(shortLinks []entity.ShortLink) []entity.ShortLink {
+	sort.SliceStable(shortLinks, func(i, j int) bool {
```
byliuyang, line 19:
```suggestion
	sort.SliceStable(shortLinks, func(firstIdx, secondIdx int) bool {
```

backend/app/usecase/search/order/created_time.go
```diff
@@ -0,0 +1,43 @@
...
+
+var _ Order = (*CreatedTime)(nil)
+
+// CreatedTime arranges searchable resources based on their created time.
+type CreatedTime struct {
+}
+
+// ArrangeShortLinks arranges shortLinks based on CreatedAt.
+// The returned short links satisfies the created before comparision if CreatedAt
+// is not nil, otherwise it sits on the end of the slices if CreatedAt is nil.
```
byliuyang, line 17:
```suggestion
```
The code seems to be very clear on the behavior of this function. Comments are usually used to explain `why` instead of `what` because everyone who reads the code can understand what but not why.

backend/app/usecase/search/order/created_time.go
```diff
@@ -0,0 +1,43 @@
...
+func (c CreatedTime) ArrangeShortLinks(shortLinks []entity.ShortLink) []entity.ShortLink {
+	sort.SliceStable(shortLinks, func(i, j int) bool {
+		if shortLinks[j].CreatedAt == nil {
+			return true
+		} else if shortLinks[i].CreatedAt == nil {
+			return false
+		}
+		return shortLinks[i].CreatedAt.Before(*shortLinks[j].CreatedAt)
```
byliuyang, line 25:
```suggestion
		iif first := shortLinks[firstIdx].CreatedAt
                 second := shortLinks[secondIdx].CreatedAt
		if first == nil && first == nil {
			return true
		}
		
		if first == nil {
		    return false
		}
		
		if second == nil {
		    return true
		}
		return first.Before(second)
```

backend/app/usecase/search/order/created_time.go
```diff
@@ -0,0 +1,43 @@
...
+// is not nil, otherwise it sits on the end of the slices if CreatedAt is nil.
+func (c CreatedTime) ArrangeUsers(users []entity.User) []entity.User {
+	sort.SliceStable(users, func(i, j int) bool {
+		if users[j].CreatedAt == nil {
+			return true
+		} else if users[i].CreatedAt == nil {
+			return false
+		}
+		return users[i].CreatedAt.Before(*users[j].CreatedAt)
+	})
```
byliuyang, line 41: I think the internal implementation of `ArrangeShortLinks` can be extracted out into a private function being reuse between itself and `ArrangeUsers`.

backend/app/usecase/search/order/created_time.go
```diff
@@ -0,0 +1,43 @@
...
+			return false
+		}
+		return shortLinks[i].CreatedAt.Before(*shortLinks[j].CreatedAt)
+	})
+	return shortLinks
+}
+
+// ArrangeUsers arranges users based on CreatedAt.
+// The returned users satisfies the created before comparision if CreatedAt
+// is not nil, otherwise it sits on the end of the slices if CreatedAt is nil.
```
byliuyang, line 32:
```suggestion
```

byliuyang:
Hi @arberiii , thank you so much for working on this cool feature! I left few comments in the PR!

backend/app/usecase/search/order/created_time.go
arberiii: Thank you for the explanation!

backend/app/usecase/search/order/created_time.go
arberiii: I agree that your suggestion is easier to read. I will modify it.

backend/app/usecase/search/order/created_time.go
arberiii: I will try to refactor!

PR #846, opened 2020-06-09
Add update Change logs functionality in ChangeLogs Repo

rohithbalaji123:
Part of #612 

### Description
Add update change logs function in sqldb adapter and change logs repository.

### DIFF
--- a/backend/app/adapter/sqldb/changelog.go
+++ b/backend/app/adapter/sqldb/changelog.go
@@ -89,6 +89,34 @@ WHERE "%s" = $1;
 	return err
 }
 
+// UpdateChange updates an existing change in change_log table
+func (c ChangeLogSQL) UpdateChange(newChange entity.Change) (entity.Change, error) {
+	statement := fmt.Sprintf(`
+UPDATE "%s"
+SET
+    "%s" = $1,
+    "%s" = $2
+WHERE "%s" = $3;
+`,
+		table.ChangeLog.TableName,
+		table.ChangeLog.ColumnTitle,
+		table.ChangeLog.ColumnSummaryMarkdown,
+		table.ChangeLog.ColumnID,
+	)
+
+	_, err := c.db.Exec(
+		statement,
+		newChange.Title,
+		newChange.SummaryMarkdown,
+		newChange.ID,
+	)
+	if err != nil {
+		return entity.Change{}, err
+	}
+
+	return newChange, nil
+}
+
 // NewChangeLogSQL creates ChangeLogSQL
 func NewChangeLogSQL(db *sql.DB) ChangeLogSQL {
 	return ChangeLogSQL{
--- a/backend/app/adapter/sqldb/changelog_integration_test.go
+++ b/backend/app/adapter/sqldb/changelog_integration_test.go
@@ -207,7 +207,7 @@ func TestChangeLogSql_DeleteChange(t *testing.T) {
 					summaryMarkdown: summaryMarkdown2,
 				},
 			},
-			deleteChangeId: "67890",
+			deleteChangeId:        "67890",
 			expectedChangeLogSize: 1,
 			expectedChangeLog: []entity.Change{
 				{
@@ -230,7 +230,7 @@ func TestChangeLogSql_DeleteChange(t *testing.T) {
 					summaryMarkdown: summaryMarkdown2,
 				},
 			},
-			deleteChangeId: "34567",
+			deleteChangeId:        "34567",
 			expectedChangeLogSize: 2,
 			expectedChangeLog: []entity.Change{
 				{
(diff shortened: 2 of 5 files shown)


backend/app/adapter/sqldb/changelog.go
```diff
@@ -89,6 +89,34 @@ WHERE "%s" = $1;
 	return err
 }
 
+// UpdateChange updates an existing change in change_log table
+func (c ChangeLogSQL) UpdateChange(newChange entity.Change) (entity.Change, error) {
+	statement := fmt.Sprintf(`
+UPDATE "%s"
+SET
```
arberiii, line 96:
```suggestion
SET "%s" = $1, "%s" = $2
```

backend/app/usecase/repository/change_log_fake.go
```diff
@@ -50,6 +50,19 @@ func (c *ChangeLogFake) removeChangeAt(idx int) error {
 	return nil
 }
 
+// UpdateChange replaces an existing change having given id with updated attributes
+func (c *ChangeLogFake) UpdateChange(newChange entity.Change) (entity.Change, error) {
+	for idx, change := range c.changeLog {
+		if change.ID == newChange.ID {
+			c.changeLog[idx].Title = newChange.Title
+			c.changeLog[idx].SummaryMarkdown = newChange.SummaryMarkdown
+			break
```
arberiii, line 59:
```suggestion
			return newChange, nil
```

backend/app/adapter/sqldb/changelog.go
```diff
@@ -89,6 +89,34 @@ WHERE "%s" = $1;
 	return err
 }
 
+// UpdateChange updates an existing change in change_log table
```
arberiii, line 92:
```suggestion
// UpdateChange updates an existing change in change_log table.
```
Forgot to add this, based on this https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

rohithbalaji123:
@byliuyang @arberiii 
I have a question regarding the value returned from the update method. Should we return back just the Change entity constructed from the user input values or should we return the updated instance from the database?
From the above implementation, the returned Change enitity will NOT have "ReleasedAt" attribute in it.

backend/app/adapter/sqldb/changelog_integration_test.go
```diff
@@ -269,6 +269,105 @@ func TestChangeLogSql_DeleteChange(t *testing.T) {
...
+		change            entity.Change
+		expectedChangeLog []entity.Change
+	}{
+		{
+			name: "update an existing change",
+			tableRows: []changeLogTableRow{
+				{
+					id:              "12345",
+					title:           "title 1",
+					summaryMarkdown: summaryMarkdown1,
```
byliuyang, line 289:
```suggestion
					summaryMarkdown: "summary 1",
```
When possible, please fill in the input data in the test case themselves directly. This will help readers focus on understanding the test case. The places we used variables were because the field require a pointer and we can only take the pointer of a variable.

arberiii:
> @byliuyang @arberiii
> I have a question regarding the value returned from the update method. Should we return back just the Change entity constructed from the user input values or should we return the updated instance from the database?
> From the above implementation, the returned Change enitity will NOT have "ReleasedAt" attribute in it.

Please look at UpdateShortLink implementation, there the result will no have "CreatedBy".
https://github.com/short-d/short/blob/71eb74d044e97a1e401931f652f36829d7d24679/backend/app/adapter/sqldb/short_link.go#L105
In my opinion, the caller should be aware only of what he requested, so the above implementation should be fine.

backend/app/adapter/sqldb/changelog.go
```diff
@@ -89,6 +89,34 @@ WHERE "%s" = $1;
...
+func (c ChangeLogSQL) UpdateChange(newChange entity.Change) (entity.Change, error) {
```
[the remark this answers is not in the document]
rohithbalaji123: Are we allowing users to update the ID field as well?
It is currently being auto-generated from the key service in create change method.

byliuyang:
@rohithbalaji123 Good question. Let's put leave as it is and put a TODO so that we can look back and revise the pattern we are using.

backend/app/adapter/sqldb/changelog_integration_test.go
rohithbalaji123: @byliuyang , we are still using the pointers to these variables below to construct `expectedChangeLog`

byliuyang: @rohithbalaji123 I see. Then fine.

backend/app/adapter/sqldb/changelog.go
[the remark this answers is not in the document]
byliuyang: Discussed offline line. We will leave the code as it is in this PR for now for the sake of simplicity.

PR #847, opened 2020-06-11
Add HasMapping to sql adapter

oatovar:
* Add ability to check if an alias exists within the user's short link repo.
* Add integration tests to cover FindAliasByUser

This is the first PR related to #798. It updates the sql adapter so that it can verify if a specific alias is owned by a given user.

### DIFF
--- a/backend/app/adapter/sqldb/user_short_link.go
+++ b/backend/app/adapter/sqldb/user_short_link.go
@@ -63,7 +63,7 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error) 
 	return aliases, nil
 }
 
-// IsAliasOwnedByUser verifies if the ShortLink alias belongs to the given user.
+// IsAliasOwnedByUser checks whether a given short link belongs to a user.
 func (u UserShortLinkSQL) IsAliasOwnedByUser(user entity.User, alias string) (bool, error) {
 	statement := fmt.Sprintf(`SELECT FROM "%s" WHERE "%s"=$1 AND "%s"=$2`,
 		table.UserShortLink.TableName,
--- a/backend/app/adapter/sqldb/user_short_link.go
+++ b/backend/app/adapter/sqldb/user_short_link.go
@@ -65,7 +65,7 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error) 
 
 // IsAliasOwnedByUser verifies if the ShortLink alias belongs to the given user.
 func (u UserShortLinkSQL) IsAliasOwnedByUser(user entity.User, alias string) (bool, error) {
-	statement := fmt.Sprintf(`SELECT from "%s" WHERE "%s"=$1 AND "%s"=$2`,
+	statement := fmt.Sprintf(`SELECT FROM "%s" WHERE "%s"=$1 AND "%s"=$2`,
 		table.UserShortLink.TableName,
 		table.UserShortLink.ColumnUserID,
 		table.UserShortLink.ColumnShortLinkAlias,
--- a/backend/app/adapter/sqldb/user_short_link.go
+++ b/backend/app/adapter/sqldb/user_short_link.go
@@ -63,7 +63,7 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error) 
 	return aliases, nil
 }
 
-// HasMapping checks whether a given short link belongs to a user.
+// HasMapping checks whether a given short link is tied to a user.
 func (u UserShortLinkSQL) HasMapping(user entity.User, alias string) (bool, error) {
 	query := fmt.Sprintf(`SELECT "%s" FROM "%s" WHERE "%s"=$1 AND "%s"=$2`,
 		table.UserShortLink.ColumnUserID,
--- a/backend/app/adapter/sqldb/user_short_link.go
+++ b/backend/app/adapter/sqldb/user_short_link.go
@@ -64,7 +64,7 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error) 
 }
 
 // IsAliasOwnedByUser checks whether a given short link belongs to a user.
-func (u UserShortLinkSQL) IsAliasOwnedByUser(user entity.User, alias string) (bool, error) {
+func (u UserShortLinkSQL) HasMapping(user entity.User, alias string) (bool, error) {
 	statement := fmt.Sprintf(`SELECT FROM "%s" WHERE "%s"=$1 AND "%s"=$2`,
 		table.UserShortLink.TableName,
 		table.UserShortLink.ColumnUserID,
--- a/backend/app/adapter/sqldb/user_short_link.go
+++ b/backend/app/adapter/sqldb/user_short_link.go
@@ -72,7 +72,7 @@ func (u UserShortLinkSQL) HasMapping(user entity.User, alias string) (bool, erro
 		table.UserShortLink.ColumnShortLinkAlias,
 	)
 
-	var id int
+	var id string
 	err := u.db.QueryRow(query, user.ID, alias).Scan(&id)
 	if err == sql.ErrNoRows {
 		return false, nil
--- a/backend/app/adapter/sqldb/user_short_link.go
+++ b/backend/app/adapter/sqldb/user_short_link.go
@@ -63,8 +63,8 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error) 
 	return aliases, nil
 }
 
-// FindAliasByUser verifies if the ShortLink alias belongs to the given user.
-func (u UserShortLinkSQL) FindAliasByUser(user entity.User, alias string) (bool, error) {
+// IsAliasOwnedByUser verifies if the ShortLink alias belongs to the given user.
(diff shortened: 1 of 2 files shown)


backend/app/adapter/sqldb/user_short_link.go
```diff
@@ -63,6 +63,24 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error)
 	return aliases, nil
 }
 
+// IsAliasOwnedByUser verifies if the ShortLink alias belongs to the given user.
+func (u UserShortLinkSQL) IsAliasOwnedByUser(user entity.User, alias string) (bool, error) {
+	statement := fmt.Sprintf(`SELECT from "%s" WHERE "%s"=$1 AND "%s"=$2`,
```
arberiii, line 68:
```suggestion
	statement := fmt.Sprintf(`SELECT FROM "%s" WHERE "%s"=$1 AND "%s"=$2`,
```

backend/app/adapter/sqldb/user_short_link_integration_test.go
```diff
@@ -114,6 +114,128 @@ func TestListShortLinkSql_FindAliasesByUser(t *testing.T) {
 	}
 }
 
+func TestListShortLinkSql_IsAliasOwnedByUser(t *testing.T) {
+	now := mustParseTime(t, "2019-05-01T08:02:16Z")
+	user := entity.User{
```
arberiii, line 119: You can directly put the user into the test struct. Also fields LastSignedInAt, CreatedAt, UpdatedAt are redundant.

backend/app/adapter/sqldb/user_short_link_integration_test.go
```diff
@@ -114,6 +114,128 @@ func TestListShortLinkSql_FindAliasesByUser(t *testing.T) {
...
+				dbMigrationRoot,
+				dbConfig,
+				func(sqlDB *sql.DB) {
+					insertUserTableRows(t, sqlDB, testCase.userTableRows)
+					insertShortLinkTableRows(t, sqlDB, testCase.shortLinkTableRows)
+					insertUserShortLinkTableRows(t, sqlDB, testCase.relationTableRows)
+
+					userShortLinkRepo := sqldb.NewUserShortLinkSQL(sqlDB)
+					result, err := userShortLinkRepo.IsAliasOwnedByUser(testCase.user, testCase.alias)
+					assert.Equal(t, nil, err)
```
arberiii, line 232: Out of context but how do you actually run these tests? I tried with scripts/integration-test but even if I put an assert.Equal(t, 1, 2) test passes

backend/app/adapter/sqldb/user_short_link_integration_test.go
oatovar: I'm a bit confused as to the reasoning behind this. Writing the user in the test struct multiple times would just create duplicated code.

oatovar: As far as the time stamps go, they're set so that there is no nil pointer left which can lead to errors down the road 😄

arberiii: I was suggesting for easy readability but we can leave it as it is since it is readable also this way. Good point about timestamps, if we change functioning, those tests should not fail for nil pointers!

byliuyang: @oatovar Making the test case easy to follow avoid distracting engineers reading the code away from coming up edge case.

backend/app/adapter/sqldb/user_short_link.go
```diff
@@ -63,6 +63,24 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error)
 	return aliases, nil
 }
 
+// IsAliasOwnedByUser verifies if the ShortLink alias belongs to the given user.
```
byliuyang, line 66:
```suggestion
// IsAliasOwnedByUser checks whether a given short link belongs to a user.
```

backend/app/adapter/sqldb/user_short_link.go
```diff
@@ -63,6 +63,24 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error)
...
+func (u UserShortLinkSQL) IsAliasOwnedByUser(user entity.User, alias string) (bool, error) {
```
byliuyang, line 67:
```suggestion
func (u UserShortLinkSQL) HasMapping(user entity.User, alias string) (bool, error) {
```

byliuyang: I prefer delegating the concept of ownership to use case.

backend/app/adapter/sqldb/user_short_link_integration_test.go
byliuyang: Let's remove the used fields and move user logic into the test case. We may have to modify the test case as the system evolve. Having user in the test case themselves allowing us to make changes without touching the rest.

backend/app/adapter/sqldb/user_short_link.go
```diff
@@ -63,6 +63,24 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error)
...
+	)
+
+	rows, err := u.db.Query(statement, user.ID, alias)
+	if err != nil {
+		return false, err
+	}
+	defer rows.Close()
+
+	found := rows.Next()
+	return found, nil
```
byliuyang, line 81: Let's try something like this so that it's more concise and readable:
```go
err := u.db.QueryRow(query, id).Scan(&id)
	if err == sql.ErrNoRows {
		return false, nil
	}
	if err != nil {
		return false, err
	}
```

backend/app/adapter/sqldb/user_short_link_integration_test.go
```diff
@@ -114,6 +114,128 @@ func TestListShortLinkSql_FindAliasesByUser(t *testing.T) {
...
+		name               string
+		userTableRows      []userTableRow
+		shortLinkTableRows []shortLinkTableRow
+		relationTableRows  []userShortLinkTableRow
+		alias              string
+		user               entity.User
+		isFound            bool
+	}{
+		{
+			name: "alias does not exist",
```
byliuyang, line 138: Let's add a case when user does not exist.

backend/app/adapter/sqldb/user_short_link_integration_test.go
```diff
@@ -114,6 +114,128 @@ func TestListShortLinkSql_FindAliasesByUser(t *testing.T) {
...
+	}
+
+	testCases := []struct {
+		name               string
+		userTableRows      []userTableRow
+		shortLinkTableRows []shortLinkTableRow
+		relationTableRows  []userShortLinkTableRow
+		alias              string
+		user               entity.User
+		isFound            bool
```
byliuyang, line 135:
```suggestion
		expectIsFound            bool
```

byliuyang: Let's add expect for the expected result so that it's intuitive for readers to figure out the purpose of the test

backend/app/adapter/sqldb/user_short_link.go
byliuyang: We may have different algorithm to determine `Ownership` in the future. This leaves us the flexibility to make changes.

backend/app/adapter/sqldb/user_short_link_integration_test.go
oatovar: I'm not sure what you mean by this. The relation table cannot have an entry without an exisiting user because of the foreign key usage. Do you mean when the user entity is not set in the test case?

byliuyang: @oatovar Awesome! Please ignore this comment then.

backend/app/adapter/sqldb/user_short_link.go
```diff
@@ -63,6 +63,26 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error)
 	return aliases, nil
 }
 
+// HasMapping checks whether a given short link belongs to a user.
```
byliuyang, line 66:
```suggestion
// HasMapping checks whether a given short link is tied to a user.
```

backend/app/adapter/sqldb/user_short_link.go
```diff
@@ -63,6 +63,26 @@ func (u UserShortLinkSQL) FindAliasesByUser(user entity.User) ([]string, error)
...
+
+	var id string
+	err := u.db.QueryRow(query, user.ID, alias).Scan(&id)
+	if err == sql.ErrNoRows {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
```
byliuyang, line 83: I wonder whether we can do the following:
```go
return true, err
```
When there is err, the return value is ignored.

oatovar: Should the value be trusted if there was an error?

oatovar: I see what you mean now. I think it would still be better to have it return as is in case someone accidentally uses the returned value.

byliuyang: @oatovar It's Go's convention not to use the returned value if there is error. I am fine keeping the code as you have.

PR #848, opened 2020-06-11
Add update Change Log Usecase and API

rohithbalaji123:
Part of #612 

### Description
Add update change log graphQL APIs along with usecase.

### DIFF
--- a/backend/app/usecase/changelog/changelog.go
+++ b/backend/app/usecase/changelog/changelog.go
@@ -18,6 +18,7 @@ type ChangeLog interface {
 	GetLastViewedAt(user entity.User) (*time.Time, error)
 	ViewChangeLog(user entity.User) (time.Time, error)
 	DeleteChange(id string) error
+	UpdateChange(id string, title string, summaryMarkdown *string) (entity.Change, error)
 }
 
 // Persist retrieves change log from and saves changes to persistent data store.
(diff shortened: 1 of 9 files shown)


backend/app/usecase/changelog/changelog.go
```diff
@@ -92,6 +93,20 @@ func (p Persist) DeleteChange(id string) error {
 	return p.changeLogRepo.DeleteChange(id)
 }
 
+// UpdateChange updates an existing change with given id in data store.
+func (p Persist) UpdateChange(
+	id string,
+	title string,
+	summaryMarkdown *string,
+) (entity.Change, error) {
+	newChange := entity.Change{
```
byliuyang, line 102: I wonder whether we should pass `user` into this use case. Every user can view the change log but only the ones with permission can update it. I think this means we need to add checks to authorizer.

backend/app/usecase/changelog/changelog_test.go
```diff
@@ -448,3 +448,111 @@ func TestPersist_DeleteChange(t *testing.T) {
...
+				tm,
+				&changeLogRepo,
+				&userChangeLogRepo,
+			)
+
+			_, err = persist.UpdateChange(testCase.change.ID, testCase.change.Title, testCase.change.SummaryMarkdown)
+			assert.Equal(t, nil, err)
+
+			changeLog, err := persist.GetChangeLog()
+
```
byliuyang, line 553:
```suggestion
```

backend/app/usecase/changelog/changelog.go
rohithbalaji123: @byliuyang , yes we should. Indeed, that is going to be my next PR. I separated it out into a new PR as we need to add authorizer to all create, update and delete usecases.

PR #849, opened 2020-06-11
Search Short Link resource

arberiii:
Search the short link with the given query from persistent storage. This will conclude this part of the search and is continuation of [PR #841](https://github.com/short-d/short/pull/841).  After retrieving all the short links for a specific user, the searcher for each short link tries to find if it matches the query in the following order:
1. Check if the alias contains the keywords by "and" operator
2. Check if the alias contains the keywords by "or" operator
3. Check if the long link contains the keywords by "and" operator
4. Check if the long link contains the keywords by "or" operator

After matches are found, it filters by limiting the number of returned results.


backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,111 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
 }
 
 func (s Search) searchShortLink(query Query, orderBy order.Order, filter Filter) (Result, error) {
-	return Result{}, nil
+	if query.User == nil {
+		return Result{}, errors.New("user not provided")
+	} else if len(query.Query) == 0 {
```
oatovar, line 74:
```suggestion
	}
	if len(query.Query) == 0 {
```

backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,112 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+
+func sortShortLinks(orderBy order.Order, shortLinkLists ...[]entity.ShortLink) {
+	for _, shortLinks := range shortLinkLists {
+		orderBy.ArrangeShortLinks(shortLinks)
+	}
+}
+
+func mergeShortLinks(shortLinkLists ...[]entity.ShortLink) []entity.ShortLink {
+	var merged []entity.ShortLink
+
```
oatovar, line 170:
```suggestion

```

backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,112 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+	if err != nil {
+		return []entity.ShortLink{}, err
+	}
+
+	return s.shortLinkRepo.GetShortLinksByAliases(aliases)
+}
+
+func matchShortLinks(shortLinks []entity.ShortLink, query Query, orderBy order.Order) []entity.ShortLink {
+	var matchedAliasByAnd, matchedAliasByOr, matchedLongLinkByAnd, matchedLongLinkByOr []entity.ShortLink
+	keywords := strings.Split(query.Query, " ")
```
oatovar, line 109: To avoid having different levels of abstraction in a single function, it would be good to move this to it's own function.
```suggestion
	keywords := getKeywordsFromQuery(query.Query)
```
Underneath, the function can just return `strings.Split(query.Query, " ")` but the calling function will not need to know about the implementation details. Not a blocker just a suggestion :smiley:

backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,112 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+		}
+
+		if strings.Contains(s, word) {
+			or = true
+		} else {
+			and = false
+		}
+	}
+	return and, or
+}
```
oatovar, line 160: This will probably need to be refactored. It returns two boolean results but it takes a slice of undetermined length which makes it a bit confusing. I'm having a bit of a hard time deciphering the and or comparisons. Could you provide a bit more context here please?

backend/app/usecase/search/search_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	for _, testCase := range testCases {
+		testCase := testCase
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+			userShortLinkRepo := repository.NewUserShortLinkRepoFake(testCase.relationUsers, testCase.relationShortLinks)
+			shortLinkRepo := repository.NewShortLinkFake(testCase.shortLinks)
+			timeout := time.Second
+
+			search := NewSearch(&shortLinkRepo, &userShortLinkRepo, timeout)
```
oatovar, line 215: This is a meta comment, but if timeouts are being used `context.Context` might be a possibility in the future.

backend/app/usecase/search/search.go
arberiii: I agree!

backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,112 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
 }
 
 func (s Search) searchShortLink(query Query, orderBy order.Order, filter Filter) (Result, error) {
-	return Result{}, nil
+	if query.User == nil {
+		return Result{}, errors.New("user not provided")
```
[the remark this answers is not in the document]
arberiii: I thought of that too, but for the moment those are the only usage! The search package will be short, so maybe we don't need to add complexity. If the error types and usage increase here then it needs to refactor.

backend/app/usecase/search/search_test.go
arberiii: Definitely! As of now, the search request is canceled only if some resource is taking a long time to return the result. In my opinion, for now, we should leave its due to its low complexity and if the complexity increases then it would be rather easy to change.

backend/app/usecase/search/search.go
arberiii: I agree! What this function does is that it takes a string say "https://google.com/" and words ["google", "facebook"]. Then it checks if the string contains all the words, which here it doesn't and simultaneously if the string contains at least one word, which in the example it does contain the word google. 

My first thought was to optimize by traversing the words only once but I see now that this would be not a good practice. Therefore I decoupled now. Please check now

backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,118 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+func (s Search) getShortLinkByUser(user entity.User) ([]entity.ShortLink, error) {
+	aliases, err := s.userShortLinkRepo.FindAliasesByUser(user)
+	if err != nil {
+		return []entity.ShortLink{}, err
+	}
+
+	return s.shortLinkRepo.GetShortLinksByAliases(aliases)
+}
+
+func matchShortLinks(shortLinks []entity.ShortLink, query Query, orderBy order.Order) []entity.ShortLink {
```
rohithbalaji123, line 107: Can we eliminate comments by adding an extra layer of methods? Feel free to disagree.
Another advantage is that, if `aliasByAnd` is true, there is no need to calculate for `aliasByOr`
For example,
```
func matchShortLinks(shortLinks []entity.ShortLink, query Query, orderBy order.Order) []entity.ShortLink {
	keywords := getKeywordsFromQuery(query.Query)
	
	for _, shortLink := range shortLinks {
		if aliasContainsAllKeywords(shortLink.Alias, keywords) {

		} else if aliasContainsAnyKeyword(shortLink.Alias, keywords) {

		} else if longLinkContainsAllKeywords(shortLink.LongLink, keywords) {

		} else if longLinkContainsAnyKeyword(shortLink.LongLink, keywords) {

		}
	}
}

func aliasContainsAllKeywords(alias string, keywords []string) {
	return stringContainsAll(alias, keywords);
}

func aliasContainsAnyKeyword(alias string, keywords []string) {
	return stringContainsAny(alias, keywords);
}

func stringContainsAll(str string, patterns []string) {
	// logic
}
```

arberiii: I do agree at some level! I think eliminating comments by adding an extra layer of methods is the right thing for readability and optimization, but I wouldn't differentiate function aliasContainsAllKeywords from longLinkContainsAllKeywords since they are the same and that's why I kept those functions under containsAllWords. Please checkout the modification after your suggestion, what do you think?

rohithbalaji123: Agreed

backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,108 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+	var matchedAliasByAll, matchedAliasByAny, matchedLongLinkByAll, matchedLongLinkByAny []entity.ShortLink
+	keywords := getKeywordsFromQuery(query.Query)
+	for _, shortLink := range shortLinks {
+		if containsAllWords(shortLink.Alias, keywords) {
```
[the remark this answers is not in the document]
byliuyang: Then match and sort can be combine together and be reused across all attributes.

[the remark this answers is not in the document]
arberiii: I do agree! Maybe we can move this interface and its implementations to the short-d/app repository because it can be applied for repositories different from Short. Wdyt?

backend/app/usecase/search/search.go
```diff
@@ -33,6 +35,7 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
 		go func() {
 			result, err := s.searchResource(filter.Resources[i], orders[i], query, filter)
 			if err != nil {
+				finalErr = err
```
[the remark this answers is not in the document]
arberiii: We can just log it then.

backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,108 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
 }
 
 func (s Search) searchShortLink(query Query, orderBy order.Order, filter Filter) (Result, error) {
-	return Result{}, nil
+	if query.User == nil {
+		return Result{}, errors.New("user not provided")
+	}
+	if len(query.Query) == 0 {
+		return Result{}, errors.New("query not provided")
```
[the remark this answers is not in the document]
arberiii: This may be not considered as an error, if a user sends an empty query then all the short links will be returned. When I was developing it in the first place, I had in mind different logic for matcher

backend/app/usecase/search/search.go
```diff
@@ -66,13 +69,108 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+		if containsAllWords(shortLink.Alias, keywords) {
+			matchedAliasByAll = append(matchedAliasByAll, shortLink)
+		} else if containsAnyWord(shortLink.Alias, keywords) {
+			matchedAliasByAny = append(matchedAliasByAny, shortLink)
+		} else if containsAllWords(shortLink.LongLink, keywords) {
+			matchedLongLinkByAll = append(matchedLongLinkByAll, shortLink)
+		} else if containsAnyWord(shortLink.LongLink, keywords) {
+			matchedLongLinkByAny = append(matchedLongLinkByAny, shortLink)
+		}
+	}
```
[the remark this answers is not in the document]
arberiii: We have to keep them together in a loop because if there is a match in the alias, then checking also for the long link may lead to duplicate results

backend/app/usecase/search/search.go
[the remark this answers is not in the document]
byliuyang: @arberiii This sound good for now.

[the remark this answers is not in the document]
byliuyang: We may apply different strategy in the future.

backend/app/usecase/search/search.go
```diff
@@ -33,6 +37,7 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
 		go func() {
 			result, err := s.searchResource(filter.Resources[i], orders[i], query, filter)
 			if err != nil {
+				s.logger.Error(err)
```
byliuyang, line 40: Let's add a TODO. We should ideally use instrumentation code instead so that we have a centralized place to organize everything and make changes: https://github.com/short-d/short/blob/master/backend/app/usecase/instrumentation/instrumentation.go.

backend/app/usecase/search/search.go
```diff
@@ -66,13 +71,85 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+		if matcher.ContainsAll(keywords, shortLink.Alias) {
+			matchedAliasByAll = append(matchedAliasByAll, shortLink)
+		} else if matcher.ContainsAny(keywords, shortLink.Alias) {
+			matchedAliasByAny = append(matchedAliasByAny, shortLink)
+		} else if matcher.ContainsAll(keywords, shortLink.LongLink) {
+			matchedLongLinkByAll = append(matchedLongLinkByAll, shortLink)
+		} else if matcher.ContainsAny(keywords, shortLink.LongLink) {
+			matchedLongLinkByAny = append(matchedLongLinkByAny, shortLink)
+		}
+	}
```
byliuyang, line 96: Let's break this into 4 separate loops for now. We can look into simplification in the future.

backend/app/usecase/search/search.go
```diff
@@ -66,13 +71,85 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+func (s Search) getShortLinkByUser(user entity.User) ([]entity.ShortLink, error) {
+	aliases, err := s.userShortLinkRepo.FindAliasesByUser(user)
+	if err != nil {
+		return []entity.ShortLink{}, err
+	}
+
+	return s.shortLinkRepo.GetShortLinksByAliases(aliases)
+}
+
+func getKeywords(query, separator string) []string {
```
byliuyang, line 128:
```suggestion
func getKeywords(query string) []string {
Doing split is an implementation detail of obtaining the keywords from the query. This can change in the future.
```

backend/app/usecase/search/search_test.go
```diff
@@ -0,0 +1,381 @@
...
+					{
+						Alias:    "git-google",
+						LongLink: "http://github.com/google",
+					},
+				},
+				users: nil,
+			},
+		},
+		{
+			name: "empty search",
```
byliuyang, line 193:
```suggestion
			name: "query no match",
```

backend/app/usecase/search/search_test.go
```diff
@@ -0,0 +1,381 @@
...
+					LongLink: "https://facebook.com",
+				},
+			},
+			expectedResult: Result{
+				shortLinks: nil,
+				users:      nil,
+			},
+		},
+		{
+			name: "one result",
```
byliuyang, line 274:
```suggestion
			name: "less matches than maxResults",
```

backend/app/usecase/search/search_test.go
```diff
@@ -0,0 +1,381 @@
...
+		name               string
+		shortLinks         shortLinks
+		Query              Query
+		filter             Filter
+		relationUsers      []entity.User
+		relationShortLinks []entity.ShortLink
+		expectedResult     Result
+	}{
+		{
+			name: "search without user",
```
byliuyang, line 29: Let's add a test case when no query is provided.

backend/app/usecase/search/search.go
```diff
@@ -66,13 +72,85 @@ func (s Search) searchResource(resource Resource, orderBy order.Order, query Que
...
+	for _, shortLink := range shortLinks {
+		if matcher.ContainsAll(keywords, shortLink.Alias) {
+			matchedAliasByAll = append(matchedAliasByAll, shortLink)
+		} else if matcher.ContainsAny(keywords, shortLink.Alias) {
+			matchedAliasByAny = append(matchedAliasByAny, shortLink)
+		} else if matcher.ContainsAll(keywords, shortLink.LongLink) {
+			matchedLongLinkByAll = append(matchedLongLinkByAll, shortLink)
+		} else if matcher.ContainsAny(keywords, shortLink.LongLink) {
+			matchedLongLinkByAny = append(matchedLongLinkByAny, shortLink)
+		}
```
byliuyang, line 96:
```suggestion
		if matcher.ContainsAll(keywords, shortLink.Alias) {
			matchedAliasByAll = append(matchedAliasByAll, shortLink)
			continue
		} 
		 if matcher.ContainsAny(keywords, shortLink.Alias) {
			matchedAliasByAny = append(matchedAliasByAny, shortLink)
			continue
		}
		 if matcher.ContainsAll(keywords, shortLink.LongLink) {
			matchedLongLinkByAll = append(matchedLongLinkByAll, shortLink)
			continue
		}
		 if matcher.ContainsAny(keywords, shortLink.LongLink) {
			matchedLongLinkByAny = append(matchedLongLinkByAny, shortLink)
		}
```

backend/app/usecase/search/search.go
byliuyang: Discussed on Slack. Postpone this to the future.

PR #842, opened 2020-06-06
Add UpdateOpenGraphMetaTags to adapter

kapillamba4:
Part of #809 

### Description
UpdateOGMetaTags sqldb adapter method returns new ShortLink after updating OpenGraph MetaTags.
A corresponding GraphQL mutation API will be created for updating OpenGraph MetaTags for a ShortLink

PR #850, opened 2020-06-11
Add UpdateTwitterTags to adapter

kapillamba4:
Part of #809 
Continuation of PR https://github.com/short-d/short/pull/842
### Description
UpdateTwitterTags sqldb adapter method returns new ShortLink after updating Twitter MetaTags.
A corresponding GraphQL mutation API will be created for updating Twitter MetaTags for a ShortLink


backend/app/adapter/sqldb/short_link.go
```diff
@@ -18,6 +18,34 @@ type ShortLinkSql struct {
 	db *sql.DB
 }
 
+// UpdateTwitterTags updates Twitter meta tags for a given short link.
+func (s *ShortLinkSql) UpdateTwitterTags(alias string, twitterTags metatag.Twitter) (entity.ShortLink, error) {
```
arberiii, line 22: Can the implementation of UpdateTwitterTags be extracted out into a private function to be used also for UpdateOpenGraphTags?

backend/app/adapter/sqldb/short_link.go
```diff
@@ -18,6 +18,34 @@ type ShortLinkSql struct {
 	db *sql.DB
 }
 
+// UpdateTwitterTags updates Twitter meta tags for a given short link.
```
arberiii, line 21: nit: the order of metatags is OpenGraph then Twitter (see backend/app/entity/metatag/metatag.go )

backend/app/adapter/sqldb/short_link.go
byliuyang: @arberiii I think we are conceptually different. Prefer keeping them separate so that we can make independent changes in the future.

backend/app/adapter/sqldb/short_link.go
byliuyang: @arberiii Sorry, didn't get what you mean. Could you please clarify? Thanks!

backend/app/adapter/sqldb/short_link.go
```diff
@@ -18,6 +18,34 @@ type ShortLinkSql struct {
...
+	)
+
+	_, err := s.db.Exec(
+		statement,
+		twitterTags.Title,
+		twitterTags.Description,
+		twitterTags.ImageURL,
+		alias,
+	)
+
```
byliuyang, line 41:
```suggestion
```

backend/app/adapter/sqldb/short_link.go
kapillamba4: How @arberiii, there Update statements are different

backend/app/adapter/sqldb/short_link.go
kapillamba4: I see

backend/app/adapter/sqldb/short_link_integration_test.go
```diff
@@ -47,6 +47,102 @@ type shortLinkTableRow struct {
...
+
+	testCases := []struct {
+		name              string
+		tableRows         []shortLinkTableRow
+		alias             string
+		metaTags          metatag.Twitter
+		expectedShortLink entity.ShortLink
+	}{
+		{
+			name: "OpenGraph tags not provided",
```
byliuyang, line 66: It seems that the name suggest those are the tests for Open Graph tags.

backend/app/adapter/sqldb/short_link.go
arberiii: If you check the order of metatags [here](https://github.com/short-d/short/blob/fix-809-part-4/backend/app/entity/metatag/metatag.go), first comes OpenGraph then Twitter. But in this PR we have UpdateTwitterTags first then UpdateOpenGraphTags

byliuyang: @arberiii Oh, okey.

backend/app/adapter/sqldb/short_link_integration_test.go
kapillamba4: @byliuyang This is similar to what we did in `TestShortLinkSql_UpdateOpenGraphTags` test cases

byliuyang: @kapillamba4 I didn’t see the value of this test case. Could you please elaborate?

kapillamba4: Resolved on Slack, will update the test cases

PR #851, opened 2020-06-13
Remove manual string concatenation for reCaptcha API request body

### LINKED ISSUE #739: [Refactor] Change ReCaptcha API body to a map
**What is frustrating you?**
With manually concatenated string, it's error prone to add new query parameter to reCaptcha.

**Your solution**
Change the manually concatenated string to a map.

Coteh:
Fixes #739.

Also added a test case for recaptcha error response. This pathway shouldn't normally happen since reCAPTCHA always returns 200 response even when there's an error, but this test case should now increase coverage of recaptcha module from 90% to 100%.

### DIFF
--- a/backend/app/adapter/recaptcha/recaptcha.go
+++ b/backend/app/adapter/recaptcha/recaptcha.go
@@ -31,7 +31,7 @@ func (r Service) Verify(captchaResponse string) (requester.VerifyResponse, error
 	apiRes := requester.VerifyResponse{}
 	err := r.http.JSON(http.MethodPost, verifyAPI, headers, body.Encode(), &apiRes)
 	if err != nil {
-		return requester.VerifyResponse{}, errors.New("Error validating captcha response")
+		return requester.VerifyResponse{}, errors.New("failed to retrieve reCaptcha API response")
 	}
 	return apiRes, nil
 }
--- a/backend/app/adapter/recaptcha/recaptcha.go
+++ b/backend/app/adapter/recaptcha/recaptcha.go
@@ -1,8 +1,9 @@
 package recaptcha
 
 import (
-	"fmt"
+	"errors"
 	"net/http"
+	"net/url"
 
 	"github.com/short-d/app/fw/webreq"
 	"github.com/short-d/short/backend/app/usecase/requester"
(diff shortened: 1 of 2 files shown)


backend/app/adapter/recaptcha/recaptcha.go
```diff
@@ -24,12 +25,13 @@ func (r Service) Verify(captchaResponse string) (requester.VerifyResponse, error
...
-	body := fmt.Sprintf("secret=%s&response=%s", r.secret, captchaResponse)
+	body := url.Values{}
+	body.Set("secret", r.secret)
+	body.Set("response", captchaResponse)
 	apiRes := requester.VerifyResponse{}
-	err := r.http.JSON(http.MethodPost, verifyAPI, headers, body, &apiRes)
+	err := r.http.JSON(http.MethodPost, verifyAPI, headers, body.Encode(), &apiRes)
 	if err != nil {
-		return requester.VerifyResponse{}, err
+		return requester.VerifyResponse{}, errors.New("Error validating captcha response")
```
byliuyang, line 34:
```suggestion
		return requester.VerifyResponse{}, errors.New("failed to get the response for reCaptcha API")
```

backend/app/adapter/recaptcha/recaptcha_integration_test.go
```diff
@@ -47,6 +49,13 @@ func TestReCaptcha_Verify(t *testing.T) {
 				Hostname:      "s.time4hacks.com",
 			},
 		},
+		{
+			name: "unsuccessful request produced an error",
```
byliuyang, line 53:
```suggestion
			name: "request failed with error",
```

backend/app/adapter/recaptcha/recaptcha_integration_test.go
```diff
@@ -47,6 +49,13 @@ func TestReCaptcha_Verify(t *testing.T) {
...
+			httpResponse: nil,
+			httpErr: errors.New("Service encountered an error"),
+			expRes: requester.VerifyResponse{},
+			expErr: errors.New("Error validating captcha response"),
```
byliuyang, line 57: Let's use has error instead of comparing error message direct.

backend/app/adapter/recaptcha/recaptcha_integration_test.go
```diff
@@ -47,6 +49,13 @@ func TestReCaptcha_Verify(t *testing.T) {
 				Hostname:      "s.time4hacks.com",
 			},
 		},
+		{
+			name: "unsuccessful request produced an error",
+			httpResponse: nil,
+			httpErr: errors.New("Service encountered an error"),
```
byliuyang, line 55:
```suggestion
			httpErr: errors.New("http request failure"),
```

backend/app/adapter/recaptcha/recaptcha_integration_test.go
Coteh: Done

PR #861, opened 2020-06-14
Refactor graphQL related schemas into a singe source

### LINKED ISSUE #860: [Refactor] Refactor graphQL related schemas into a singe source
**What is frustrating you?**
There are multiple places where the GraphQL schema is being reused in services like `ChangeLogGraphQLApi`, `ShortLinkGraphQLApi` and `UrlService`. When there is a change in the server graphQL schema, it requires changes in all these files which makes it harder to maintain.

**Your solution**
Create a single source for all these schema and make the service import and use the interface structures.

rohithbalaji123:
Resolves #860

### Description
Add multiple copies of AuthQuery, AuthMutation, Query, Mutation in each GraphQL Service classes. If the graphQL schema changes, it requires changing the schema in all the files and the code is quite redundant.

### Description
Separate out short graphql schema structure in a separate file and use that structure wherever required.


byliuyang:
We also need to remember to rename URL to short links during the fixit week.

rohithbalaji123:
@byliuyang , please don't merge yet. We need to rename Schema to schema in all files. I'll do a sanity test once updated and merge it.

PR #864, opened 2020-06-15
Add match package

arberiii:
When searching for short links we need to check if the keywords match an alias or a long link. Decoupling search package with the logic of keyword matching has two main properties of best practice: readability and reusability. 

The motivation of this PR comes from [the on hold PR #849](https://github.com/short-d/short/pull/849).

Initially, I was going to add match package to the app repo but decided to put it here.


backend/app/usecase/match/contains_all.go
```diff
@@ -0,0 +1,25 @@
...
+
+// ContainsAll checks if an input contains all the elements of a list.
+type ContainsAll struct {
+}
+
+// IsMatch checks if the input contains all the words.
+func (c ContainsAll) IsMatch(words []string, input string) bool {
+	if input == "" {
+		return false
+	} else if len(words) == 0 {
```
rohithbalaji123, line 15:
```suggestion
	}
	if len(words) == 0 {
```

backend/app/usecase/match/contains_all.go
```diff
@@ -0,0 +1,25 @@
...
+
+var _ Keyword = (*ContainsAll)(nil)
+
+// ContainsAll checks if an input contains all the elements of a list.
+type ContainsAll struct {
+}
+
+// IsMatch checks if the input contains all the words.
+func (c ContainsAll) IsMatch(words []string, input string) bool {
+	if input == "" {
```
rohithbalaji123, line 13: Shouldn't `IsMatch([""], "")` return `true`?

rohithbalaji123: I guess, we can remove this condition. Empty string is still a valid string and if I am not wrong, `strings.Contains(input, word)` should handle this edge case by itself.

backend/app/usecase/match/contains_all.go
```diff
@@ -0,0 +1,25 @@
...
+		return false
+	} else if len(words) == 0 {
+		return false
```
rohithbalaji123, line 16: Logically speaking, shouldn't this be `true`?
`input` has all the items(in this case, no items) given in the `words`.
Again, this case is handled by the main logic below.

backend/app/usecase/match/contains_any.go
```diff
@@ -0,0 +1,25 @@
...
+var _ Keyword = (*ContainsAny)(nil)
+
+// ContainsAny checks if an input contains any element of a list.
+type ContainsAny struct {
+}
+
+// IsMatch checks if the input contains any word.
+func (c ContainsAny) IsMatch(words []string, input string) bool {
+	if input == "" {
+		return false
```
rohithbalaji123, line 14: Again, `IsMatch([""], "")` should return `true` IMO.

backend/app/usecase/match/contains_any.go
```diff
@@ -0,0 +1,25 @@
...
+	} else if len(words) == 0 {
+		return false
```
rohithbalaji123, line 16: This here is correct, the `input` does not have any patterns given in the `words`. Again, this is also handled by the main logic below.

backend/app/usecase/match/keyword.go
```diff
@@ -0,0 +1,32 @@
+package match
+
+import (
+	"fmt"
+)
+
+// KeywordType represents keyword matching type.
+type KeywordType int
```
rohithbalaji123, line 8: Actually, it is not Keyword type. It is a type of matching algorithm. Can we name this something else to make it obvious?

backend/app/usecase/match/keyword.go
```diff
@@ -0,0 +1,32 @@
...
+
+const (
+	// ContainsAllKeywords represents ContainsAll type
+	ContainsAllKeywords KeywordType = iota
+	// ContainsAnyKeyword represents ContainsAny type
+	ContainsAnyKeyword
+)
+
+// Keyword matches a list of words against an input.
+type Keyword interface {
```
rohithbalaji123, line 18: Just curious if `Keyword` is the right term. Shouldn't it have some sort of `Matcher` in the name?

backend/app/usecase/match/contains_all.go
arberiii: That is true, my bad. Thank you very much for pointing it out.

backend/app/usecase/match/contains_all.go
arberiii: You are right! I had in mind the search package and to optimize there I was thinking to reject if they are empty. Let me fix this now

backend/app/usecase/match/keyword.go
arberiii: The package name is match, so the term here is match.Keyword. If we name it something like match.KeywordMatcher it seems too much unnecessary information. I agree however that it is a bit hard to follow. I will modify it now.

backend/app/usecase/match/keyword.go
arberiii: How about changing to Algorithm?

rohithbalaji123: Sounds good to me. Yeah, `match.Algorithm` sounds better than `match.KeywordType`

backend/app/usecase/match/keyword.go
rohithbalaji123: Okay, understood.

backend/app/usecase/match/contains_any.go
```diff
@@ -0,0 +1,22 @@
...
+
+// ContainsAny checks if an input contains any element of a list.
+type ContainsAny struct {
+}
+
+// IsMatch checks if the input contains any word.
+func (c ContainsAny) IsMatch(words []string, input string) bool {
+	if len(words) == 0 {
+		return true
+	}
```
rohithbalaji123, line 15:
```suggestion
```
In my previous comment, I meant the implementation was right. This should be false. And this is already handled by the logic below.
Does the `input` contain any words from the `words` list? No(when the `words` list is empty).

backend/app/usecase/match/contains_all_test.go
```diff
@@ -0,0 +1,72 @@
...
+			name:     "no match",
+			words:    []string{"a", "b", "c"},
+			input:    "xyz",
+			expected: false,
+		},
+		{
+			name:     "in between match",
+			words:    []string{"a", "b", "c"},
+			input:    "xcz",
+			expected: false,
```
rohithbalaji123, line 58:
```suggestion
			name:     "at least one mismatch",
			words:    []string{"a", "b", "c"},
			input:    "xcz",
			expected: false,
```

backend/app/usecase/match/contains_any_test.go
```diff
@@ -0,0 +1,72 @@
...
+			expected: true,
+		},
+		{
+			name:     "no match",
+			words:    []string{"a", "b", "c"},
+			input:    "xyz",
+			expected: false,
+		},
+		{
+			name:     "in between match",
```
rohithbalaji123, line 55:
```suggestion
			name:     "any one match",
```

backend/app/usecase/match/contains_all.go
```diff
@@ -0,0 +1,19 @@
+package match
+
+import "strings"
+
+var _ Keyword = (*ContainsAll)(nil)
+
+// ContainsAll checks if an input contains all the elements of a list.
```
byliuyang, line 7:
```suggestion
// ContainsAll checks whether the input contains all the elements in the list.
```

backend/app/usecase/match/contains_all.go
```diff
@@ -0,0 +1,19 @@
...
+
+// IsMatch checks if the input contains all the words.
+func (c ContainsAll) IsMatch(words []string, input string) bool {
+	for _, word := range words {
+		if !strings.Contains(input, word) {
+			return false
+		}
+	}
+	return true
+}
```
byliuyang, line 19: It seems that this doesn't have any dependencies. Can this be a standalone function instead?

backend/app/usecase/match/contains_any.go
```diff
@@ -0,0 +1,19 @@
+package match
+
+import "strings"
+
+var _ Keyword = (*ContainsAny)(nil)
+
+// ContainsAny checks if an input contains any element of a list.
+type ContainsAny struct {
+}
```
byliuyang, line 9: So did here.

backend/app/usecase/match/keyword.go
byliuyang: Let's rename the package to matcher. We want to reuse this package for future input validation also. If both contains all and contain any are changed to standalone function, this file is not needed any more.

backend/app/usecase/match/contains_all.go
arberiii: I agree!

backend/app/usecase/match/keyword.go
arberiii: I had it in the begging named matcher, but was in dilemma 😅

backend/app/usecase/match/contains_all.go
byliuyang: Please remove the structs and just export the functions!

Issue #869, opened 2020-06-20
[BUG] Runtime error: Index out of range when orders length is smaller than resources in Search


arberiii:
**Describe the bug**
Application crashes when there are fewer orders then resources when the Search function is called. This happens because of this code (written by me): 
https://github.com/short-d/short/blob/0b0b8a0114e92540a143f5ee43d81caca527e0ae/backend/app/usecase/search/search.go#L34-L38

**To Reproduce**
1. Create an instance of type Search 
2. Call Search function on it, with Filter parameter that has Resource = []Resource{ShortLink} and no orders.

**Expected behavior**
If the order is not given then the order of that resource should be Unchanged.

PR #870, opened 2020-06-20
Add constructor for Filter

arberiii:
Fixes [issue #869](https://github.com/short-d/short/issues/869).

### DIFF
--- a/backend/app/usecase/search/filter.go
+++ b/backend/app/usecase/search/filter.go
@@ -10,14 +10,23 @@ const (
 	User
 )
 
-// OrderedResource represents a type of searchable objects together with the order.
-type OrderedResource struct {
-	Resource Resource
-	Order    order.By
-}
-
 // Filter represents the filters for a search request.
 type Filter struct {
-	MaxResults       int
-	OrderedResources []OrderedResource
+	MaxResults int
+	Resources  []Resource
+	Orders     []order.By
+}
+
+// IsValid checks if the resources and orders have one-to-one relation.
+func (f *Filter) IsValid() bool {
+	return len(f.Resources) == len(f.Orders)
+}
+
+// NewFilter creates Filter.
+func NewFilter(maxResults int, resources []Resource, orders []order.By) Filter {
+	return Filter{
+		MaxResults: maxResults,
+		Resources:  resources,
+		Orders:     orders,
+	}
 }
--- a/backend/app/usecase/search/filter.go
+++ b/backend/app/usecase/search/filter.go
@@ -10,9 +10,14 @@ const (
 	User
 )
 
+// OrderedResource represents a type of searchable objects together with the order.
+type OrderedResource struct {
+	Resource Resource
+	Order    order.By
+}
+
 // Filter represents the filters for a search request.
 type Filter struct {
-	MaxResults int
-	Resources  []Resource
-	Orders     []order.By
+	MaxResults       int
+	OrderedResources []OrderedResource
 }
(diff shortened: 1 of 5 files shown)


backend/app/usecase/search/search.go
```diff
@@ -31,6 +31,9 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
 	resultCh := make(chan Result)
 	defer close(resultCh)
 
+	if len(filter.Resources) > len(filter.Orders) {
+		filter.Orders = appendDefaultOrders(filter.Orders, len(filter.Resources)-len(filter.Orders))
```
rohithbalaji123, line 35: @arberiii , just curious how does this maintain the one to one correspondence of the resource and order? Should we think about extracting resource and order into a separate structure in [Filter](https://github.com/short-d/short/blob/master/backend/app/usecase/search/filter.go).

arberiii: I agree that your suggestion seems better. One mismatch as you mention offline is if the package's client supplies 3 resources and 2 order by's for first and third resource.

backend/app/usecase/search/filter.go
```diff
@@ -10,9 +10,14 @@ const (
 	User
 )
 
+// OrderedResource represents a type of searchable objects together with the order.
+type OrderedResource struct {
+	Resource Resource
+	Order    order.By
+}
```
byliuyang, line 17: I like the idea of combining them into a single struct. However, it does look weird especially when we are adding 1 more attribute in the future. Can we add `func NewFilter() (Filter, error)` and `(f *F) IsValid() bool` and return error in search function instead?

arberiii: Okay, let me implement your suggestion then we can compare all three alternatives

backend/app/usecase/search/filter.go
```diff
@@ -16,3 +16,17 @@ type Filter struct {
...
 	Orders     []order.By
 }
+
+// IsValid checks if the resources and orders have one-to-one relation.
+func (f *Filter) IsValid() bool {
+	return len(f.Resources) == len(f.Orders)
+}
+
+// NewFilter creates Filter.
+func NewFilter(maxResults int, resources []Resource, orders []order.By) Filter {
```
byliuyang, line 26: We should throw error here is there is mismatch between resource and order

backend/app/usecase/search/search.go
```diff
@@ -28,6 +28,10 @@ type Result struct {
 
 // Search finds resources based on specified criteria.
 func (s Search) Search(query Query, filter Filter) (Result, error) {
+	if !filter.IsValid() {
+		return Result{}, errors.New("filter is not valid")
```
byliuyang, line 32:
```suggestion
		return Result{}, errors.New("invalid filter")
```

backend/app/usecase/search/filter_test.go
```diff
@@ -0,0 +1,65 @@
+package search
+
+import (
+	"testing"
+
+	"github.com/short-d/app/fw/assert"
+	"github.com/short-d/short/backend/app/usecase/search/order"
+)
+
+func TestCreatedTime_ArrangeShortLinks(t *testing.T) {
```
byliuyang, line 10:
```suggestion
func TestNewFilter(t *testing.T) {
```

arberiii: used it as template 😓 . Thanks for the catch.

PR #875, opened 2020-06-22
Remove unnecessary dereference operator

byliuyang:
It seems that DB mutations are working as properly as expected. For the sake of consistency in `wire.go`, removing dereference operator from the receivers.


backend/app/adapter/sqldb/short_link.go
```diff
@@ -11,15 +11,15 @@ import (
 	"github.com/short-d/short/backend/app/usecase/repository"
 )
 
-var _ repository.ShortLink = (*ShortLinkSql)(nil)
+var _ repository.ShortLink = (*ShortLinkSQL)(nil)
 
-// ShortLinkSql accesses ShortLink information in short_link table through SQL.
-type ShortLinkSql struct {
+// NewShortLinkSQL accesses ShortLink information in short_link table through SQL.
```
arberiii, line 16:
```suggestion
// ShortLinkSQL accesses ShortLink information in short_link table through SQL.
```

PR #876, opened 2020-06-23
Add update short link api

byliuyang:
### Description
Currently, once a short link is created, there is way to edit it. This cause significant inconvenience when user accidentally created a undesired short link using their favorite alias. 

### Description
Add update short link API to allow user to edit short link they created.


byliuyang:
Reviewed by myself.

PR #878, opened 2020-06-23
Read search API request

arberiii:
Previously we developed the route for search API and in this PR we read the search request. First, we start by authenticating the user and then we decode the JSON request to get the query and the filter.


backend/app/adapter/routing/handle/search.go
```diff
@@ -1,17 +1,97 @@
...
 ) router.Handle {
 	return func(w http.ResponseWriter, r *http.Request, params router.Params) {
-		w.Write([]byte("not implemented"))
+		user := getUser(r, authenticator)
+		if user == nil {
+			w.Write([]byte("user not authenticated"))
+			return
+		}
+
+		searchRequest, err := readSearchRequest(r)
```
byliuyang, line 26:
```suggestion
		searchBody, err := parseSearchBody(r)
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -1,17 +1,97 @@
...
+	b, err := ioutil.ReadAll(r.Body)
+	defer r.Body.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	var searchRequest SearchRequest
+	err = json.Unmarshal(b, &searchRequest)
+	return &searchRequest, err
+}
```
byliuyang, line 69: Please considering defining custom unmarshal function instead of letting API consumers invoke parse functions manually: https://gist.github.com/odeke-em/a2f0ad491c5d097e1ea29ecd036644a5

backend/app/adapter/routing/handle/search.go
```diff
@@ -1,17 +1,97 @@
...
+	"io/ioutil"
 	"net/http"
 
 	"github.com/short-d/app/fw/router"
+	"github.com/short-d/short/backend/app/usecase/authenticator"
 	"github.com/short-d/short/backend/app/usecase/search"
+	"github.com/short-d/short/backend/app/usecase/search/order"
 )
 
 // Search fetches resources under certain criterias.
```
byliuyang, line 14:
```suggestion
// Search fetches resources under certain criteria.
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -1,17 +1,118 @@
...
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (s *SearchRequest) UnmarshalJSON(data []byte) error {
+	temp := struct {
+		Query  string `json:"query"`
+		Filter struct {
+			MaxResults int      `json:"max_results"`
+			Resources  []string `json:"resources"`
+			Orders     []string `json:"orders"`
+		} `json:"filter"`
```
byliuyang, line 52: Let's create a separate `Unmarshal` for Filter. We can specify `resources` and `orders` as enums define in the use case because we are going to convert the type within `Filter`'s `Unmarshal` function.

backend/app/adapter/routing/handle/search.go
```diff
@@ -1,17 +1,169 @@
...
+
+func (s *SearchRequest) setFilter(maxResults int, resources []search.Resource, orders []order.By) error {
+	filter, err := search.NewFilter(maxResults, resources, orders)
+	if err != nil {
+		return err
+	}
+
+	s.filter = filter
+	return nil
+}
```
byliuyang, line 81: I didn't extra value provided by having those 2 function. It doesn't seems that they are going to change frequently. Let's move them into `UnmarshalJSON` instead. Thanks!

backend/app/adapter/routing/handle/search.go
```diff
@@ -1,17 +1,169 @@
...
+	temp := struct {
+		Query  string `json:"query"`
+		Filter Filter `json:"filter"`
+	}{}
+
+	if err := json.Unmarshal(data, &temp); err != nil {
+		return err
+	}
+
+	s.setQuery(temp.Query)
```
byliuyang, line 63: I think both of those are parsed automatically and we don't have to explicitly `UnmarshalJSON` for `SearchRequest`. Let's put `json` tag on the request and filter themselves.

PR #881, opened 2020-06-25
Effectual assignments in Go code and constant comments 

arberiii:
We have some ineffectual assignments in Go code as reported in [Go Report Card](https://goreportcard.com/report/github.com/short-d/short). It is easy to fix by simply handling the error. I also checked for ineffectual assignments after the fix and there are none.

In the [previous PR](https://github.com/short-d/short/pull/878) I commented on the block of constants but it didn't add much value. I marked and modified them but please feel free to suggest completely different comments for them.

### DIFF
--- a/backend/app/usecase/search/filter.go
+++ b/backend/app/usecase/search/filter.go
@@ -9,10 +9,12 @@ import (
 // Resource represents a type of searchable objects.
 type Resource uint
 
-// Resource constants represent supported types of resource.
 const (
+	// Unknown represent the resource that is not listed or known.
 	Unknown Resource = iota
+	// ShortLink represents the short link resource.
 	ShortLink
+	// User represent the user resource.
 	User
 )
 
--- a/backend/app/usecase/search/order/order.go
+++ b/backend/app/usecase/search/order/order.go
@@ -5,9 +5,10 @@ import "github.com/short-d/short/backend/app/entity"
 // By represents the order of resources being sorted in.
 type By uint
 
-// By constants represent supported types of order.
 const (
+	// ByUnsorted represent the order type which does not sort the resource.
 	ByUnsorted By = iota
+	// ByCreatedTimeASC represent the order type which sorts resources based on created time.
 	ByCreatedTimeASC
 )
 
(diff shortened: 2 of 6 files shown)


backend/app/usecase/search/filter.go
```diff
@@ -9,10 +9,12 @@ import (
 // Resource represents a type of searchable objects.
 type Resource uint
 
-// Resource constants represent supported types of resource.
 const (
+	// Unknown represents the resource that is not listed or known.
```
byliuyang, line 13:
```suggestion
	// Unknown implies the resource type is not support by the search module. 
	// This is usually used as the fallback resource by the callers of search API.
```

backend/app/usecase/search/filter.go
```diff
@@ -9,10 +9,12 @@ import (
...
 	Unknown Resource = iota
+	// ShortLink represents the short link resource.
```
byliuyang, line 15:
```suggestion
	// ShortLink represents the short links.
```

backend/app/usecase/search/filter.go
```diff
@@ -9,10 +9,12 @@ import (
...
 	ShortLink
+	// User represents the user resource.
```
byliuyang, line 17:
```suggestion
	// User represents the internal user.
```

backend/app/usecase/search/order/order.go
```diff
@@ -5,9 +5,10 @@ import "github.com/short-d/short/backend/app/entity"
 // By represents the order of resources being sorted in.
 type By uint
 
-// By constants represent supported types of order.
 const (
+	// ByUnsorted represents the order type which does not sort the resource.
```
byliuyang, line 9:
```suggestion
	// ByUnsorted keeps the original order of search results.
```

backend/app/usecase/search/order/order.go
```diff
@@ -5,9 +5,10 @@ import "github.com/short-d/short/backend/app/entity"
...
 	ByUnsorted By = iota
+	// ByCreatedTimeASC represents the order type which sorts resources based on created time.
```
byliuyang, line 11:
```suggestion
	// ByCreatedTimeASC sorts search results based on their creation time.
```

PR #882, opened 2020-06-25
Response to the Search API request

arberiii:
This PR concludes the Search API feature. Since for now, we allow users to search only short links created by them, we first authenticate the user. After that, we retrieve the result from the searcher. The results are formatted into JSON format and the API responds to the client.

### DIFF
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -6,9 +6,8 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/short-d/short/backend/app/entity"
-
 	"github.com/short-d/app/fw/router"
+	"github.com/short-d/short/backend/app/entity"
 	"github.com/short-d/short/backend/app/usecase/authenticator"
 	"github.com/short-d/short/backend/app/usecase/search"
 	"github.com/short-d/short/backend/app/usecase/search/order"
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -13,6 +13,8 @@ import (
 	"github.com/short-d/short/backend/app/usecase/search/order"
 )
 
+const maxResults = 100
+
 var searchResource = map[string]search.Resource{
 	"short_link": search.ShortLink,
 	"user":       search.User,
@@ -35,8 +37,8 @@ type Filter struct {
 	Orders     []order.By
 }
 
-// SearchRespond represents the respond to the Search API request.
-type SearchRespond search.Result
+// SearchResponse represents the response to the Search API request.
+type SearchResponse search.Result
 
 // ShortLink represents the short_link field of Search API respond.
 type ShortLink entity.ShortLink
@@ -69,6 +71,9 @@ func Search(
 			Query: body.Query,
 			User:  user,
 		}
+		if body.Filter.MaxResults == 0 {
+			body.Filter.MaxResults = maxResults
+		}
 		filter, err := search.NewFilter(body.Filter.MaxResults, body.Filter.Resources, body.Filter.Orders)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
@@ -81,8 +86,8 @@ func Search(
 			return
 		}
 
-		respond := SearchRespond(results)
-		respBody, err := json.Marshal(&respond)
+		response := SearchResponse(results)
+		respBody, err := json.Marshal(&response)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
@@ -127,7 +132,7 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
 }
 
 // MarshalJSON formats the search result into json format.
-func (s *SearchRespond) MarshalJSON() ([]byte, error) {
+func (s *SearchResponse) MarshalJSON() ([]byte, error) {
 	shortLinks := make([]ShortLink, len(s.ShortLinks))
 	for i := 0; i < len(s.ShortLinks); i++ {
 		shortLinks[i] = ShortLink(s.ShortLinks[i])
@@ -138,11 +143,11 @@ func (s *SearchRespond) MarshalJSON() ([]byte, error) {
 	}
 
 	buf := struct {
-		ShortLink []ShortLink `json:"short_link,omitempty"`
-		User      []User      `json:"user,omitempty"`
+		ShortLinks []ShortLink `json:"short_links,omitempty"`
+		Users      []User      `json:"users,omitempty"`
 	}{
-		ShortLink: shortLinks,
-		User:      users,
+		ShortLinks: shortLinks,
+		Users:      users,
 	}
 
 	return json.Marshal(&buf)
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -2,11 +2,12 @@ package handle
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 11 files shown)


backend/app/adapter/routing/handle/search.go
```diff
@@ -34,9 +35,18 @@ type Filter struct {
 	Orders     []order.By
 }
 
+// SearchRespond represents the respond to the Search API request.
+type SearchRespond search.Result
```
rohithbalaji123, line 39:
```suggestion
// SearchResponse represents the response to the Search API request.
type SearchResponse search.Result
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -54,7 +64,31 @@ func Search(
...
+			return
+		}
+
+		results, err := searcher.Search(query, filter)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		respond := SearchRespond(results)
```
rohithbalaji123, line 84:
```suggestion
		response := SearchResponse(results)
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -54,7 +64,31 @@ func Search(
 			return
 		}
 
-		w.Write([]byte(fmt.Sprintf("%v", body)))
+		user := getUser(r, authenticator)
+		query := search.Query{
+			Query: body.Query,
+			User:  user,
+		}
+		filter, err := search.NewFilter(body.Filter.MaxResults, body.Filter.Resources, body.Filter.Orders)
```
rohithbalaji123, line 72: Is `MaxResults` a mandatory attribute? What happens if it not passed in search query? Will it be set to a defined default value or is it going to zero based on Go's default int value?

backend/app/adapter/routing/handle/search.go
```diff
@@ -91,3 +125,65 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
...
+		users[i] = User(s.Users[i])
+	}
+
+	buf := struct {
+		ShortLink []ShortLink `json:"short_link,omitempty"`
+		User      []User      `json:"user,omitempty"`
+	}{
+		ShortLink: shortLinks,
+		User:      users,
+	}
```
rohithbalaji123, line 146:
```suggestion
	buf := struct {
		ShortLinks []ShortLink `json:"short_links,omitempty"`
		Users      []User      `json:"users,omitempty"`
	}{
		ShortLinks: shortLinks,
		Users:      users,
	}
```

backend/app/adapter/routing/handle/search.go
arberiii: you are correct! I will modify it now so we have a constant of max results in the case that it is not passed in the search query

byliuyang: @rohithbalaji123 To me I think the correct the behavior should be when maxResult not specified, we return all.

byliuyang: I think this part of business logic instead of API level thing.

backend/app/adapter/routing/handle/search.go
```diff
@@ -34,9 +37,18 @@ type Filter struct {
 	Orders     []order.By
 }
 
+// SearchResponse represents the response to the Search API request.
+type SearchResponse search.Result
```
byliuyang, line 41: I wonder whether the structure of search response can be directly specified here so that developers know what fields are expected. We may need a private constructor to convert `result` to `search response`.

backend/app/adapter/routing/handle/search.go
```diff
@@ -34,9 +37,18 @@ type Filter struct {
...
+
+// ShortLink represents the short_link field of Search API respond.
+type ShortLink entity.ShortLink
```
byliuyang, line 44: Ditto

backend/app/adapter/routing/handle/search.go
arberiii: Good point!

backend/app/adapter/routing/handle/search.go
arberiii: I will modify it and we can compare

backend/app/adapter/routing/handle/search.go
```diff
@@ -54,7 +80,31 @@ func Search(
...
+			return
+		}
+
+		results, err := searcher.Search(query, filter)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		response := resultToResponse(results)
```
byliuyang, line 100:
```suggestion
		response := newSearchResponse(results)
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -91,3 +141,40 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
 	}
 	return nil
 }
+
+func resultToResponse(result search.Result) SearchResponse {
```
byliuyang, line 145:
```suggestion
func newSearchResponse(result search.Result) SearchResponse {
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -91,3 +141,40 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
...
+	shortLinks := make([]ShortLink, len(result.ShortLinks))
+	for i := 0; i < len(result.ShortLinks); i++ {
+		shortLinks[i] = convertShortLink(result.ShortLinks[i])
```
byliuyang, line 148:
```suggestion
		shortLinks[i] = newShortLink(result.ShortLinks[i])
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -91,3 +141,40 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
...
+	}
+	users := make([]User, len(result.Users))
```
byliuyang, line 150:
```suggestion

	users := make([]User, len(result.Users))
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -91,3 +141,40 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
...
+	for i := 0; i < len(result.Users); i++ {
+		users[i] = convertUser(result.Users[i])
```
byliuyang, line 152:
```suggestion
		users[i] = newUser(result.Users[i])
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -91,3 +141,40 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
...
+	}
+
+	return SearchResponse{
+		ShortLinks: shortLinks,
+		Users:      users,
+	}
+}
+
+func convertShortLink(shortLink entity.ShortLink) ShortLink {
```
byliuyang, line 161:
```suggestion
func newShortLink(shortLink entity.ShortLink) ShortLink {
```

backend/app/adapter/routing/handle/search.go
```diff
@@ -91,3 +141,40 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
...
+	return ShortLink{
+		Alias:     shortLink.Alias,
+		LongLink:  shortLink.LongLink,
+		ExpireAt:  shortLink.ExpireAt,
+		CreatedAt: shortLink.CreatedAt,
+		UpdatedAt: shortLink.UpdatedAt,
+	}
+}
+
+func convertUser(user entity.User) User {
```
byliuyang, line 171:
```suggestion
func newUser(user entity.User) User {
```

PR #885, opened 2020-06-25
Add instrumentation for Search API

arberiii:
This PR adds instrumentation to the search API. It has only two metrics: search-failed and search-succeded.

### DIFF
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -2,11 +2,12 @@ package handle
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
+	"time"
 
 	"github.com/short-d/app/fw/router"
+	"github.com/short-d/short/backend/app/entity"
 	"github.com/short-d/short/backend/app/usecase/authenticator"
 	"github.com/short-d/short/backend/app/usecase/search"
 	"github.com/short-d/short/backend/app/usecase/search/order"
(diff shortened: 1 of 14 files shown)


backend/app/adapter/routing/handle/search.go
```diff
@@ -87,24 +93,28 @@ func Search(
...
 		response := newSearchResponse(results)
 		respBody, err := json.Marshal(&response)
 		if err != nil {
+			i.SearchFailed(err)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
 
 		w.Write(respBody)
+		i.SearchSucceed()
```
byliuyang, line 117: We need to track the keyword being searched, in which city and at what time so that we can analyze and determine the new features we want to build in order to better serve the users. This means we need to include analytics.

backend/app/usecase/instrumentation/instrumentation.go
```diff
@@ -91,6 +94,29 @@ func (i Instrumentation) FeatureToggleRetrievalFailed(err error) {
...
+func (i Instrumentation) SearchSucceed(keywords string, t time.Time) {
+	go func() {
+		c := <-i.searchSucceedCh
+		i.metrics.Count("search-succeed", 1, 1, c)
+		userID := i.getUserID(nil)
+		props := map[string]string{
+			"keywords": keywords,
+			"time":     t.String(),
+		}
+		i.analytics.Track("Search", props, userID, c)
```
arberiii, line 107: @byliuyang can you please check why tracking location is not working? You can see in Segment dashboard that the city field is empty also for MadeFeatureDecision event. However, the keywords field is working fine.

My guess is that the proxy server is not correct. It seems that header X-Forwarded-For should be set manually from client in order to work.

byliuyang: @arberiii It doesn't work locally. We have to test it on testing branch. Sorry about this.

arberiii: No worries, thank you for info!

backend/app/usecase/instrumentation/instrumentation.go
```diff
@@ -91,6 +94,29 @@ func (i Instrumentation) FeatureToggleRetrievalFailed(err error) {
...
 
+// SearchSucceed tracks the successes when searching the short resources.
+func (i Instrumentation) SearchSucceed(user *entity.User, keywords string, t time.Time) {
+	go func() {
+		c := <-i.searchSucceedCh
+		i.metrics.Count("search-succeed", 1, 1, c)
+		userID := i.getUserID(user)
+		props := map[string]string{
+			"keywords": keywords,
+			"time":     t.String(),
```
byliuyang, line 105: Time is automatically tracked by analytics API. I wonder whether we can track the list of `resources` being search also. Let's separate them with `,`.

backend/app/usecase/instrumentation/instrumentation.go
```diff
@@ -91,6 +94,29 @@ func (i Instrumentation) FeatureToggleRetrievalFailed(err error) {
...
 
+// SearchSucceed tracks the successes when searching the short resources.
+func (i Instrumentation) SearchSucceed(user *entity.User, keywords string, resources []string) {
+	go func() {
+		c := <-i.searchSucceedCh
+		i.metrics.Count("search-succeed", 1, 1, c)
+		userID := i.getUserID(user)
+		props := map[string]string{
+			"keywords":  keywords,
+			"resources": strings.Join(resources, ", "),
```
byliuyang, line 105:
```suggestion
			"resources": strings.Join(resources, ","),
```

PR #900, opened 2020-06-27
Add create change service APIs

rohithbalaji123:
Part of #612 

### Description
Add service methods to create a new change. <del>I have tried to split the `createChange` method in `ChangeGraphQLApi` into multiple methods with each focusing on only one thing. Please confirm if the approach is alright.</del> => Reverting back to previous style.


frontend/src/entity/Change.tsx
```diff
@@ -1,6 +1,17 @@
...
 }
+
+export function parseChange(gqlChange: IShortGraphQLChange): Change {
+  return {
+    id: gqlChange.id,
+    title: gqlChange.title,
+    summaryMarkdown: gqlChange.summaryMarkdown,
+    releasedAt: new Date(gqlChange.releasedAt)
+  };
+}
```
byliuyang, line 17: I wonder whether they can move this to GraphQL package. Entities are core business logic and shouldn't depends on concrete technologies.

byliuyang: Entities are the most stable and used everywhere.

frontend/src/service/Change.service.ts
```diff
@@ -0,0 +1,25 @@
+import { Err, ErrorService } from './Error.service';
```
byliuyang, line 1: I wonder whether we should combine this file with change log API. It seems that they can be grouped into a single unit.

frontend/src/service/Error.service.ts
```diff
@@ -70,6 +71,13 @@ const networkErr = {
               `
 };
 
+const unauthorizedErr = {
+  name: 'Unauthorized error',
+  description: `
+              You are not authorized to perform this action.
```
byliuyang, line 77:
```suggestion
              You don't have permission needed to perform this action.
```

frontend/src/service/shortGraphQL/ChangeGraphQL.api.ts
```diff
@@ -0,0 +1,71 @@
...
+  private async loadCreateChangeGqlVariables(
+    title: string,
+    summary: string
+  ): Promise<any> {
+    return {
+      captchaResponse: await this.captchaService.execute(CREATE_CHANGE),
+      authToken: this.authService.getAuthToken(),
+      change: { title: title, summaryMarkdown: summary }
+    };
+  }
```
byliuyang, line 70: I wonder whether we can put them in a single function. Doesn't seems to necessary because we are already inside API service. What are your thoughts on this?

frontend/src/service/Change.service.ts
```diff
@@ -0,0 +1,25 @@
...
+  ) {}
+
+  createChange(title: string, summaryMarkdown: string): Promise<Change> {
+    return new Promise<Change>((resolve, reject) => {
+      this.changeGraphQLApi
+        .createChange(title, summaryMarkdown)
+        .then(resolve)
+        .catch(errCode => {
+          if (errCode === Err.Unauthenticated) {
+            reject({ authenticationErr: 'User is not authenticated' });
```
byliuyang, line 18: I wonder whether we should hard code the string here. What are their purpose?

rohithbalaji123: @byliuyang This was the format being used in the [ShortLinks service](https://github.com/short-d/short/blob/master/frontend/src/service/Url.service.ts#L73). Should be try to refactor to a generic one everywhere?

byliuyang: @rohithbalaji123 Let’s refactor & clean up everything!

rohithbalaji123: @byliuyang , we should come up with a definite error handling mechanism. The lines [here](https://github.com/short-d/short/blob/master/frontend/src/service/Url.service.ts#L71) and [here](https://github.com/short-d/short/blob/master/frontend/src/service/ShortLink.service.ts#L30) handles error in a way, whereas [this](https://github.com/short-d/short/blob/master/frontend/src/service/ShortLink.service.ts#L42) in a different way. Which are we finalizing on?

byliuyang: @rohithbalaji123 Can we add a TODO here so that we look into it later without blocking this PR?

Issue #595, opened 2020-03-22
[Feature] Limit number of updates shown for users.


rohithbalaji123:
**What is the problem?**
When showing changes logs in the website, there is a possibility that a user might login after a really long time when he will he fed with too many changes making it difficult for the user to go through all the changes.

**Your solution**
We can limit the number of logs shown to an appropriate count(n). Also, we can't just choose the last n updates as the user might miss some major updates that happened earlier. For this purpose, we can add a field called `isMajorUpdate` in `change_log` table, which will be set if we think it's a major update and users should definitely know about it.

So, the condition will be, fetch latest n number of updates along with major ones since he last visited the site. We could also improve to have the final logs match the count(n) by fetching less of minor updates.

**Additional context**
If feasible, we could make carousal for such updates to have GIFs which explains the usage of that update.

byliuyang:
Love this!

PR #905, opened 2020-06-28
Add Change getter graphQL api

rohithbalaji123:
Part of #612 

### Description
Add `changes` graphQL API to fetch all the changes available to be displayed in admin panel. This is different from `changeLog` in the sense that, `changeLog` also fetches `lastViewedAt` timestamp which is not related to admin panel functionalities. The newly added API internally reuses the same repository methods used for `changeLog`. This also adds the permission check whether the user can view the changes.


backend/app/usecase/changelog/changelog_test.go
```diff
@@ -448,6 +448,116 @@ func TestPersist_ViewChangeLog(t *testing.T) {
...
+				},
+				{
+					ID:              "54321",
+					Title:           "title 2",
+					SummaryMarkdown: &summaryMarkdown2,
+				},
+			},
+			availableKeys: []keygen.Key{},
+			roles: map[string][]role.Role{
+				"alpha": {role.Admin},
```
arberiii, line 494:
```suggestion
				"alpha": {role.ViewChange},
```
I think is better to test with the user that has the specific role

backend/app/adapter/gqlapi/resolver/authquery.go
```diff
@@ -55,6 +55,26 @@ func (v AuthQuery) ChangeLog() (ChangeLog, error) {
 	return newChangeLog(changeLog, lastViewedAt), err
 }
 
+// Changes retrieves all the changes that exists in the persistent storage
```
arberiii, line 58:
```suggestion
// Changes retrieves all the changes that exists in the persistent storage.
```
nit: The comments in go should end with a full stop.

byliuyang:
@rohithbalaji123 GraphQL client side query can decide which fields to fetch and ignore those that are not needed.

backend/app/adapter/gqlapi/schema.go
```diff
@@ -47,6 +47,9 @@ type AuthQuery {
 	"""Fetch the all the changes visible to the current user"""
 	changeLog: ChangeLog!
 
+	"""Fetch all the changes that exists in the system"""
+	changes: [Change!]!
```
byliuyang, line 51:
```suggestion
	allChanges: [Change!]!
```

backend/app/usecase/changelog/changelog.go
```diff
@@ -27,10 +27,11 @@ func (e ErrUnauthorizedAction) Error() string {
 
 // ChangeLog retrieves change log and create changes.
 type ChangeLog interface {
-	CreateChange(title string, summaryMarkdown *string, user entity.User) (entity.Change, error)
 	GetChangeLog() ([]entity.Change, error)
 	GetLastViewedAt(user entity.User) (*time.Time, error)
 	ViewChangeLog(user entity.User) (time.Time, error)
+	CreateChange(title string, summaryMarkdown *string, user entity.User) (entity.Change, error)
+	GetChanges(user entity.User) ([]entity.Change, error)
```
byliuyang, line 34:
```suggestion
	GetAllChanges(user entity.User) ([]entity.Change, error)
```

rohithbalaji123:
The reason why I went for a new API is that they are being used for completely different purposes and their fetching mechanism might change in future.
For instance, when we start implementing #595 , that will be specific to user side changelogs, while admin will still need to get all the changes.

PR #910, opened 2020-06-29
Append Emotic to all component styles to avoid style leaking

byliuyang:
### Description
There is a chance to accidentally modify the styles of Emotic when embedding it on external site due to CSS naming conflicts.

### Description
Prefix all Emotic specific styles with `Emotic` to avoid styling leaking & conflicts.


byliuyang:
I will revise the PR descriptions later

PR #913, opened 2020-06-29
Add update change client service methods

rohithbalaji123:
Part of #612 

### Description
Add update change client service methods which allows authorized user to update existing changes.

### DIFF
--- a/frontend/src/service/ChangeLog.service.ts
+++ b/frontend/src/service/ChangeLog.service.ts
@@ -90,6 +90,22 @@ export class ChangeLogService {
     });
   }
 
+  updateChange(id: string, change: Change): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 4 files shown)


frontend/src/service/ChangeLog.service.ts
```diff
@@ -110,7 +126,7 @@ export class ChangeLogService {
     return new Promise(async (resolve, reject) => {
       this.changeLogGraphQLApi
         .getAllChanges()
-        .then(resolve)
+        .then(changes => resolve(this.sortChanges(changes)))
```
byliuyang, line 129: I wonder why this is needed?

rohithbalaji123: @byliuyang The `getAllChanges` is not returning changes in the same order when some updates happen. It's better if we follow some intiutive order while showing for the admins to moderate.

PR #919, opened 2020-07-01
Handle duplicated alias in Usecase

arberiii:
### Description
When a user tries to update a short link but the updated short link alias already exists then it fails because of the Postgres database unique on the alias.
### Screenshots
image

### Description
Handle error in Usecase, which is independent of storage and more friendly. Before pushing the commit, I ran the format script, which added some changes.

### DIFF
--- a/backend/app/usecase/shortlink/updater_test.go
+++ b/backend/app/usecase/shortlink/updater_test.go
@@ -95,6 +95,47 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
 			expectedShortLink: entity.ShortLink{},
 		},
 		{
+			name:  "update alias already exist",
+			alias: "git",
+			shortlinks: shortLinks{
+				"git": entity.ShortLink{
+					Alias:     "git",
+					LongLink:  "https://github.com/short-d",
+					UpdatedAt: &now,
+				},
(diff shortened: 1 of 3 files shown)


backend/app/usecase/shortlink/updater_test.go
```diff
@@ -94,6 +94,47 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
 			expectedHasErr:    true,
 			expectedShortLink: entity.ShortLink{},
 		},
+		{
+			name:  "update alias already exist",
```
byliuyang, line 98:
```suggestion
			name:  "alias already exist",
```

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -94,6 +94,47 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
...
+					UpdatedAt: &now,
+				},
+				{
+					Alias:     "short-d",
+					LongLink:  "http://short-d.com/",
+					UpdatedAt: &now,
+				},
+			},
+			expectedHasErr:    true,
+			expectedShortLink: entity.ShortLink{},
```
byliuyang, line 136:
```suggestion
```

Issue #823, opened 2020-05-31
[Refactor] Migrate switch statement for error type checking to user errors.As instead


Coteh:
**What is frustrating you?**
When short was first created, error types were checked using switch statement. Since then, golang has added two new constructs to work with errors: `Is` and `As`. Can be read about here:
https://blog.golang.org/go1.13-errors

Error type checking is still done the old way in the following three instances:
- `authmutation.go` [1](https://github.com/short-d/short/blob/a900268c57539d040500f31b8f01ec406e002cc4/backend/app/adapter/gqlapi/resolver/authmutation.go#L66-L77)
- `changelog.go` [1](https://github.com/short-d/short/blob/a900268c57539d040500f31b8f01ec406e002cc4/backend/app/usecase/changelog/changelog.go#L58-L61) [2](https://github.com/short-d/short/blob/a900268c57539d040500f31b8f01ec406e002cc4/backend/app/usecase/changelog/changelog.go#L74-L82)

**Your solution**
Since we're checking for error types, `errors.As` should be used to check the error type against each of the different error types.

**Alternatives considered**
Can also use `errors.Is` to check for sentinel error values, but like [this article](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) points out, sentinel errors are not very flexible as even a change of context for the error will fail the check.
The same article also suggests opaque errors, which allow errors to be checked by its behaviour instead of its type. This can also be considered as an alternative over using `errors.As` but will require a greater refactor effort.

PR #920, opened 2020-07-01
Refactor error type checking to use errors.As

Coteh:
Changed the convention of error type checking to use the new style introduced in Go 1.13, addressing #823.

https://blog.golang.org/go1.13-errors

One of the major benefits of using this style is that `errors.As` (and `errors.Is`) will examine the entire error chain and give you the error if it's within the chain. A drawback is that the syntax looks a bit more verbose than switch statements, but the ability to examine the error chain makes up for it I think!

### DIFF
--- a/backend/app/adapter/gqlapi/resolver/authmutation.go
+++ b/backend/app/adapter/gqlapi/resolver/authmutation.go
@@ -1,6 +1,7 @@
 package resolver
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


backend/app/adapter/gqlapi/resolver/authmutation.go
```diff
@@ -90,19 +91,22 @@ func (a AuthMutation) CreateShortLink(args *CreateShortLinkArgs) (*ShortLink, er
...
-	case shortlink.ErrInvalidLongLink:
-		return nil, ErrInvalidLongLink{u.LongLink, string(err.(shortlink.ErrInvalidLongLink).Violation)}
-	case shortlink.ErrInvalidCustomAlias:
-		return nil, ErrInvalidCustomAlias{*customAlias, string(err.(shortlink.ErrInvalidCustomAlias).Violation)}
-	case shortlink.ErrMaliciousLongLink:
+	} else if errors.As(err, &l) {
+		return nil, ErrInvalidLongLink{u.LongLink, string(l.Violation)}
+	} else if errors.As(err, &c) {
+		return nil, ErrInvalidCustomAlias{*customAlias, string(c.Violation)}
+	} else if errors.As(err, &m) {
```
byliuyang, line 106:
```suggestion
	}
	if errors.As(err, &l) {
		return nil, ErrInvalidLongLink{u.LongLink, string(l.Violation)}
	}
	if errors.As(err, &c) {
		return nil, ErrInvalidCustomAlias{*customAlias, string(c.Violation)}
	}
	if errors.As(err, &m) {
```

byliuyang: It seems that the else statements are unnecessary and would interferer readers from understanding the code.

byliuyang:
Hi @Coteh , thank you so much for resolving this tech debt! Please run `./scripts/format` to fix codeclimate comment!

PR #921, opened 2020-07-02
Fix and add test cases for short link updater

arberiii:
The test case "reject malicious long link" has an error because the short link is not founded and not because it is malicious. This PR fixes that and adds a test case when the alias is not valid.


backend/app/usecase/shortlink/updater_test.go
```diff
@@ -164,6 +164,36 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
 			expectedHasErr:    true,
 			expectedShortLink: entity.ShortLink{},
 		},
+		{
+			name:  "alias is invalid",
```
byliuyang, line 168:
```suggestion
			name:  "alias contains hash tag",
```

PR #924, opened 2020-07-02
Fix long link auto fill in bug

byliuyang:
### Description
Long link was not automatically filled in when user clicks on the chrome extension on their favorite web page. This bug is introduced during https://github.com/short-d/short/pull/887 refactor.

Text field originally always uses external state as it's input value. When the long link is attached in the URL, the long link state in the parent is updated. After the refactor, the text field only uses its own internal state and ignored the updated external value.

### Description
Invoke `updateValue` on the TextField to force update it's value when automatically filling in the long link. This PR also refactored home page component and moved short link creation logic into `CreateShortLinkSection`.

### DIFF
--- a/frontend/src/component/form/TextField.tsx
+++ b/frontend/src/component/form/TextField.tsx
@@ -31,6 +31,12 @@ export class TextField extends Component<IProps, IState> {
     this.props.onChange(text);
   };
 
+  updateValue(text: string) {
+    this.setState({
+      text: text
+    });
+  }
+
   handleBlur = () => {
     if (!this.props.onBlur) {
       return;
--- a/frontend/src/component/form/TextField.tsx
+++ b/frontend/src/component/form/TextField.tsx
@@ -2,40 +2,36 @@ import React, { ChangeEvent, Component, createRef } from 'react';
 import './TextField.scss';
 
 interface IProps {
-  defaultText?: string;
+  text?: string;
   placeHolder?: string;
   onChange?: (text: string) => void;
   onBlur?: () => void;
 }
 
-interface IState {
-  text: string;
-}
-
-export class TextField extends Component<IProps, IState> {
+export class TextField extends Component<IProps, any> {
   textInput = createRef<HTMLInputElement>();
 
-  constructor(props: IProps) {
-    super(props);
-    this.state = {
-      text: props.defaultText || ''
-    };
-  }
+  render = () => {
+    return (
+      <input
+        ref={this.textInput}
+        className={'text-field'}
+        type={'text'}
+        value={this.props.text}
+        onChange={this.handleChange}
+        onBlur={this.handleBlur}
+        placeholder={this.props.placeHolder}
+      />
+    );
+  };
 
   handleChange = (event: ChangeEvent<HTMLInputElement>) => {
     const text = event.target.value;
-    this.setState({ text: text });
     if (!this.props.onChange) {
       return;
     }
     this.props.onChange(text);
   };
-
-  updateValue(text: string) {
-    this.setState({
-      text: text
-    });
-  }
 
   handleBlur = () => {
     if (!this.props.onBlur) {
@@ -46,19 +42,5 @@ export class TextField extends Component<IProps, IState> {
 
   focus = () => {
     this.textInput.current!.focus();
-  };
-
-  render = () => {
-    return (
-      <input
-        ref={this.textInput}
-        className={'text-field'}
-        type={'text'}
-        value={this.state.text}
-        onChange={this.handleChange}
-        onBlur={this.handleBlur}
-        placeholder={this.props.placeHolder}
-      />
-    );
   };
 }
--- a/frontend/src/component/pages/HomePage.tsx
+++ b/frontend/src/component/pages/HomePage.tsx
@@ -6,8 +6,6 @@ import { Url } from '../../entity/Url';
 import { Footer } from './shared/Footer';
 import { SignInModal } from './shared/sign-in/SignInModal';
 import { ExtPromo } from './shared/promos/ExtPromo';
-import { validateLongLinkFormat } from '../../validators/LongLink.validator';
-import { validateCustomAliasFormat } from '../../validators/CustomAlias.validator';
 import { Location, History } from 'history';
 import { AuthService } from '../../service/Auth.service';
 import { IBrowserExtensionService } from '../../service/extensionService/BrowserExtension.service';
@@ -17,13 +15,8 @@ import { UIFactory } from '../UIFactory';
 import { IAppState } from '../../state/reducers';
 import { Store, Unsubscribe } from 'redux';
 import {
-  raiseCreateShortLinkError,
   raiseGetUserShortLinksError,
-  raiseGetChangeLogError,
-  raiseInputError,
-  updateAlias,
-  updateCreatedUrl,
-  updateLongLink
+  raiseGetChangeLogError
 } from '../../state/actions';
 import { ErrorService } from '../../service/Error.service';
 import { UrlService } from '../../service/Url.service';
@@ -65,13 +58,6 @@ interface State {
   isUserSignedIn?: boolean;
   shouldShowAdminButton?: boolean;
   shouldShowPromo?: boolean;
-  longLink?: string;
-  alias?: string;
-  shortLink?: string;
-  createdUrl?: Url;
-  qrCodeUrl?: string;
-  inputErr?: string;
-  isShortLinkPublic?: boolean;
   autoCompleteSuggestions?: Array<Url>;
   changeLog?: Array<Change>;
   currentPagedShortLinks?: IPagedShortLinks;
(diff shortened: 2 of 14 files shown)


frontend/src/state/actions.ts
```diff
@@ -10,7 +9,6 @@ export const UPDATE_LONG_LINK = 'UPDATE_LONG_URL';
 export const UPDATE_ALIAS = 'UPDATE_ALIAS';
```
byliuyang, line 9: Need to remove unused reducer!

frontend/src/component/pages/HomePage.tsx
```diff
@@ -345,65 +308,12 @@ export class HomePage extends Component<Props, State> {
...
-        <ErrorModal store={this.props.store} />
+  private handleOnShortLinkCreated = (shortLink: string) => {
+    this.copyShortenedLink(shortLink);
+    this.refreshUserShortLinks();
+  };
 
-        <Toast ref={this.toastRef} />
-      </div>
-    );
+  private handleOnRequestSignIn = () => {
```
rohithbalaji123, line 316: Do you think we should rename this to `authenticationErrHandler`? We will let the home page decide what happens when the user is not authenticated.

frontend/src/component/pages/shared/CreateShortLinkSection.tsx
```diff
@@ -7,77 +7,188 @@ import { ShortLinkUsage } from './ShortLinkUsage';
...
-  onPublicToggleClick?: (enabled: boolean) => void;
-  onCreateShortLinkButtonClick?: () => void;
+  shouldShowUsage: boolean;
+  createdShortLink: string;
+  createdLongLink: string;
+  qrCodeURL: string;
 }
 
-export class CreateShortLinkSection extends Component<Props> {
+export class CreateShortLinkSection extends Component<IProps, IState> {
```
rohithbalaji123, line 38: This is no longer a presentational component :)

frontend/src/component/pages/shared/CreateShortLinkSection.tsx
```diff
@@ -7,77 +7,188 @@ import { ShortLinkUsage } from './ShortLinkUsage';
...
 
-interface Props {
+interface IProps {
+  store: Store<IAppState>;
   uiFactory: UIFactory;
-  longLinkText?: string;
+  urlService: UrlService;
+  qrCodeService: QrCodeService;
+  onShortLinkCreated?: (shortLink: string) => void;
+  onRequestSignIn?: () => void;
```
rohithbalaji123, line 24: I would prefer it being named as `authenticationErrHandler`. Let the parent component decide what to do.

frontend/src/component/form/TextField.tsx
```diff
@@ -31,6 +31,12 @@ export class TextField extends Component<IProps, IState> {
     this.props.onChange(text);
   };
 
+  updateValue(text: string) {
```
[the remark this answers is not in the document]
rohithbalaji123: Actually, why not add both defaultText and text and make them both optional props? I am guessing, that's how it is being done internally in `<input>`. The key is to make sure the `text` prop is updated as and when the value changes in `InputField`. Without `changeHandler`, it will keep resetting the value.
```
changeHandler(event) {
    this.setState({ value: event.target.value });
}

<TextField text={this.state.value} onChange={this.changeHandler} />
```

frontend/src/component/pages/shared/CreateShortLinkSection.tsx
```diff
@@ -7,77 +7,188 @@ import { ShortLinkUsage } from './ShortLinkUsage';
...
+}
+
+interface IState {
+  longLink: string;
   alias?: string;
-  shortLink?: string;
-  inputErr?: string;
-  createdUrl?: Url;
-  qrCodeUrl?: string;
+  inputError: string | null;
```
rohithbalaji123, line 30: nit: 
TypeScript team doesn't use null : [TypeScript coding guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined).
Douglas Crockford thinks [null is a bad idea](https://www.youtube.com/watch?v=PSGEjv3Tqo0&feature=youtu.be&t=9m21s)

frontend/src/component/pages/shared/CreateShortLinkSection.tsx
```diff
@@ -7,77 +7,188 @@ import { ShortLinkUsage } from './ShortLinkUsage';
...
   isShortLinkPublic?: boolean;
-  onLongLinkTextFieldBlur?: () => void;
-  onLongLinkTextFieldChange?: (newLongLink: string) => void;
-  onShortLinkTextFieldBlur?: () => void;
-  onShortLinkTextFieldChange?: (newAlias: string) => void;
-  onPublicToggleClick?: (enabled: boolean) => void;
-  onCreateShortLinkButtonClick?: () => void;
+  shouldShowUsage: boolean;
```
rohithbalaji123, line 32: Just curious if we can make this a function instead. One less state to maintain and it is dependent on the existence of the below 3 states.
```
private shouldShowUsage() {
    return (this.state.createdLongLink && this.state.createdShortLink && this.state.qrCodeURL);
}
```

frontend/src/component/pages/shared/CreateShortLinkSection.tsx
```diff
@@ -7,77 +7,188 @@ import { ShortLinkUsage } from './ShortLinkUsage';
...
           </div>
         )}
       </Section>
     );
   }
+
+  autoFillInLongLink(longLink: string) {
+    if (validateLongLinkFormat(longLink) != null) {
+      return;
+    }
```
rohithbalaji123, line 104: Is this needed here? We can just show the error when the user blurs out of the `InputField`.
For instance, if the long link is too long, the user has the option to make it smaller by removing any unwanted fragments(hashed fragments maybe) in the link.

frontend/src/component/form/TextField.tsx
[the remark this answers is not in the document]
byliuyang: @rohithbalaji123 I recommend not to add text prop for now. I am trying to avoid magical value update when possible because that's usually a source of bug. The data should flow in one direction without having to guess what's going on.

frontend/src/component/pages/HomePage.tsx
byliuyang: Done.

frontend/src/component/pages/shared/CreateShortLinkSection.tsx
byliuyang: Done.

frontend/src/component/pages/shared/CreateShortLinkSection.tsx
byliuyang: Done.

frontend/src/component/pages/shared/CreateShortLinkSection.tsx
rohithbalaji123: Okay, in that case, can we raise input error here instead of dropping it altogether?

byliuyang: @rohithbalaji123 Sounds good. Will do.

frontend/src/component/form/TextField.tsx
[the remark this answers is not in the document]
rohithbalaji123: @byliuyang 
Actually, this introduces two sources of truth for the value contained in the input field. By having this API, we are quite literally adding a way to update the state which the prop already does for us. Isn't it the whole purpose of props?

Now, coming back to the two sources of truth issue, consider a case where onChange listener looks something like this,
```
onChangeValue(value: string) {
    if (isInvalid(value))
        return;
    this.setState({ value });
}
```
In the above scenario, the text field will show one thing and the state which holds the value for usage is something else. I feel having the text as a prop is a more natural and changing that will need developers to keep in mind about this intricacy.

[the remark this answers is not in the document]
byliuyang: @rohithbalaji123 I was thinking the value is an internal state of text field. The only way for the outside world to change it is explicitly invoking update API so that we know exactly where we are updating it's value when reading the code.  I was very confused where did update the long link when doing the previous refactor and introduced this bug. Want to avoid this from happening again.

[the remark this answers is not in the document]
rohithbalaji123: @byliuyang , Okay, in that case, the ideal solution would have been to have a method called `updateLongLink` which sets the state of `longLink`. That seems more like a React way.
```
updateLongLink(to: string) {
    this.setState({ longLink: to });
}
```
https://reactjs.org/docs/refs-and-the-dom.html#when-to-use-refs

Anyways, this method still logically works. I am ready to mark it resolved.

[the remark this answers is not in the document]
byliuyang: @rohithbalaji123 Okey. I can change this.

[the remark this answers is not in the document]
byliuyang: @rohithbalaji123 made changes as you suggested as reading the React doc.

PR #927, opened 2020-07-03
Error type check in shortlink and validator packages

arberiii:
Test cases that expect errors usually pass if it has any error, which is not correct if the error is different from what is expected. This PR introduces error checking in test tables.


backend/app/usecase/shortlink/retriever_test.go
```diff
@@ -110,7 +115,7 @@ func TestRetriever_GetShortLink(t *testing.T) {
 			shortLink, err := retriever.GetShortLink(testCase.alias, testCase.expiringAt)
 
 			if testCase.hasErr {
-				assert.NotEqual(t, nil, err)
+				assert.Equal(t, testCase.err, err)
```
byliuyang, line 118: In Golang, it's the best practice not to check the equality of error suggested both by Dave Chaney and stackoverflow in unit tests: https://www.reddit.com/r/golang/comments/680s4j/using_custom_error_types_for_testing. If we really want to, let's check error types using `error.As` instead of checking equality. The error string was only designed for human readability.

byliuyang: It's very likely that we changed the string for better readability and suddenly many tests start failing.

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -285,7 +293,7 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
 
 			shortLink, err := updater.UpdateShortLink(testCase.alias, testCase.update, testCase.user)
 			if testCase.expectedHasErr {
-				assert.NotEqual(t, nil, err)
+				assert.Equal(t, true, errors.As(err, &testCase.err))
```
byliuyang, line 296: This checks they are both `Error` types but didn't verify what error it is.

Coteh: I'd argue that only the type of error matters for these test cases. It doesn't necessarily matter what the value inside the errors are, and checking the value in addition to the type could make tests more brittle.

arberiii: @byliuyang isn't that what we want. Ultimately we would like to have a specific error type for each error.

byliuyang: @Coteh @arberiii I wasn't sure whether the error interface in the test case structure will prevent from verifying the real error type. I wonder whether we can experiment with it and see whether this work when we assign the concrete error  to the error interface.

byliuyang: Sorry for not being clear.

arberiii: It is weird since I'm testing with testCase.err = ErrMaliciousLongLink and err  = ErrShortLinkNotFound, and the tests still passes

arberiii: I'm not sure why but something weird is happening with errors.As function. If I create two errors like this:
```go
err1 = ErrShortLinkNotFound
err2 = ErrMaliciousLongLink("HTTP")
errors.As(err1, &err2) // false
```
But if we have:
```go
err = ErrShortLinkNotFound
testCase.err = ErrMaliciousLongLink("HTTP")
errors.As(err2, &testCase.err) // true
```
Any idea what is happening?

arberiii:
For the moment, we need a better, cleaner and more accurate solution for checking error type while testing. This feature needs more time with research and design, so I’m closing it for now.

PR #928, opened 2020-07-05
Provide fake reCaptcha verifier in dev

### LINKED ISSUE #790: [Feature] Disable reCaptcha in Dev Environment
**What is the problem?**
Testing any authenticated content in the dev environment require updating recaptcha to always return true. When testing in a dev environment recaptcha, should automatically be disabled so that contributors do not modify this part of the code during local testing.

**Your solution**
Enable/Disable features depending on the environment of the application.

Coteh:
Fixes #790. Adds a provider that provides the real `Verifier` (The reCaptcha version that has always been there) in non-dev environments and a fake version that always returns true in dev environment. This should remove the need to comment out the verifier in `mutation.go` while working on the project.


backend/app/usecase/requester/verifier.go
```diff
@@ -1,22 +1,27 @@
 package requester
 
-// Verifier verifies in coming network to prevents cyber attacks.
-type Verifier struct {
+// Verifier verifies incoming network to prevent cyber attacks.
+type Verifier interface {
+	IsHuman(recaptchaResponse string) (bool, error)
+}
+
+// VerifierReCaptcha verifies incoming network using ReCaptcha to prevent cyber attacks.
```
Coteh, line 8: I capitalized reCaptcha in the doc comments, but should it be lowercase? Or does it not matter?

backend/app/usecase/requester/verifier_fake.go
```diff
@@ -0,0 +1,14 @@
+package requester
+
+// VerifierFake is a stub of Verifier to be used in development only.
+type VerifierFake struct {}
+
+// IsHuman checks whether the request is sent by a human user.
```
Coteh, line 6: I can probably change this comment to be more reflective of its fake nature.

Coteh: Other fake method comments in the codebase don't seem to be changed so I'm gonna leave it.

backend/app/usecase/requester/verifier.go
Coteh: I see the r capitalized in other comments so gonna leave it for now.

backend/app/usecase/requester/verifier.go
```diff
@@ -1,22 +1,27 @@
...
+type VerifierReCaptcha struct {
```
arberiii, line 9: We have to make sure that VerifierReCaptcha satisfies Verifier interface. It can be done simply by:
```
var _  Verifier = (*VerifierReCaptcha)(nil)
```

Also in my opinion interface and implementation should be separated. Please look at usecase/repository where are most of the interfaces with fake implementation.

Coteh: Thanks for the feedback! Just made the fixes and pushed, wasn't sure where to put the ReCaptcha implementation of `Verifier` so I left it in the same package as the interface (requester).

backend/app/adapter/gqlapi/api_test.go
```diff
@@ -64,7 +64,7 @@ func TestGraphQlAPI(t *testing.T) {
 	)
 
 	s := requester.NewReCaptchaFake(requester.VerifyResponse{})
-	verifier := requester.NewVerifier(s)
+	verifier := requester.NewVerifierReCaptcha(s)
```
byliuyang, line 67:
```suggestion
	verifier := requester. NewReCaptchaVerifier(s)
```

backend/app/usecase/requester/verifier_fake.go
```diff
@@ -0,0 +1,14 @@
+package requester
+
+// VerifierFake is a stub of Verifier to be used in development only.
```
byliuyang, line 3:
```suggestion
// VerifierFake turns off human verification during local development.
```

backend/app/usecase/requester/verifier_recaptcha.go
```diff
@@ -0,0 +1,24 @@
+package requester
+
+// VerifierReCaptcha verifies incoming network using ReCaptcha to prevent cyber attacks.
```
byliuyang, line 3:
```suggestion
// VerifierReCaptcha verifies incoming network using ReCaptcha to prevent spamming attacks.
```

backend/app/usecase/requester/verifier_recaptcha.go
```diff
@@ -0,0 +1,24 @@
...
+type VerifierReCaptcha struct {
+	service ReCaptcha
+}
+
+var _ Verifier = (*VerifierReCaptcha)(nil)
```
byliuyang, line 8: Let's put this before type definition so that it's easier to reader to know the functionality of struct before reading into the details

byliuyang:
Hi @Coteh , thank you so much for working on this cool feature! Let's few comments on the PR! Overall looks good!

PR #934, opened 2020-07-05
Add close icon option to modal

rohithbalaji123:
### Description
Whenever a modal is defined, we add a close icon in all the places with almost the same properties. So, similar code is repeated in multiple places.

### Description
Move the close icon into the modal component and add a flag to let the client component decide whether or not to show the close icon.


byliuyang:
Huh. This would force all modals to have the same close button. The close button for the `error` modal and the change logs have different styles.

rohithbalaji123:
@byliuyang , yes that's right. That gives us a couple of advantages.
1. Uniformity of the Modal is maintained throughout the project.
1. For simple color changes like in the case of Error Modal can be done through a single line CSS change like [here](https://github.com/short-d/short/pull/934/files#diff-54c90eba5d9df149735171f704ccb4ffR31) instead of rewriting the whole CSS.

If there is too much of a difference, which ideally shouldn't happen within a project, then the component can choose to set `showCloseIcon` flag to `false` and build their own Icon.

byliuyang:
@rohithbalaji123 Prefer composition over flags based on open close principal: Open to extension, close to modification.

byliuyang:
Modal contents have different color & styles. The close button has to adapt to the color theme of the content. The styling diverge from here.

PR #935, opened 2020-07-05
Add create change section

rohithbalaji123:
Part of #612 

### Description
Add create change logs section to allow admins to create a new change log.

### Screenshots
image


byliuyang:
Let's create a new director called `admin/changelog` inside `pages` directory and move all relation components to there! Thanks

frontend/src/component/pages/admin/changelog/CreateChangeSection.scss
```diff
@@ -0,0 +1,48 @@
...
+      flex-direction: row;
+      flex-wrap: wrap;
+      margin-top: 20px;
+      justify-content: space-between;
+
+      .markdown-viewer {
+        padding: 0 10px;
+        background-color: $color-create-change-markdown-background;
+        overflow-y: auto;
+      }
```
byliuyang, line 24: If the internal class name of text field or markdown viewer changed, this will break. Let's try not rely on the internal implementation of those elements.
I recommend wrapping text field and markdown viewer with div element and style that element instead. We also need to modify those element internally and make make them adapt to the width & height of the potential parent.

frontend/src/component/pages/admin/changelog/CreateChangeSection.scss
```diff
@@ -0,0 +1,48 @@
...
+        box-sizing: border-box;
+        height: 250px;
+      }
+
+      @include respond-to($phone) {
+        .text-area,
+        .markdown-viewer {
+          width: 100%;
+        }
+      }
```
byliuyang, line 38: Ditto

frontend/src/component/pages/admin/changelog/CreateChangeSection.scss
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    .button {
+      background-color: $color-medium-green;
+      padding: 15px 50px;
+      width: fit-content;
+      margin: 30px auto 0;
+    }
```
byliuyang, line 46: Let's edit button's props so that we can configure styling for it.

frontend/src/component/pages/admin/changelog/CreateChangeSection.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  render() {
+    return (
+      <div className="create-change-section">
+        <Section title="Create New Change">
+          <div className="create-change-form">
+            <TextField onChange={this.handleTitleChange} placeHolder="Title" />
+            <div className="change-summary">
+              <TextAreaField
+                onChange={this.handleSummaryChange}
```
byliuyang, line 42:
```suggestion
                onChange={this.handleOnSummaryChange}
```

byliuyang:
It's probably the time we should look into theming.

frontend/src/component/pages/admin/changelog/CreateChangeSection.scss
byliuyang: How about this: Let's add a TODO for theming & CSS module. I'll start looking into them this week. Then we can plan for large scale refactor.

frontend/src/component/pages/admin/changelog/CreateChangeSection.scss
byliuyang: Let's add a TODO for so that we don't forget when doing styling.

byliuyang: We now support CSS module. This can be done by supporting a new style.

frontend/src/component/pages/admin/changelog/CreateChangeSection.scss
byliuyang: We now support CSS module. This can be done by supporting a new style.

Issue #744, opened 2020-05-06
[Refactor] Disallow special URL characters as aliases


Coteh:
**What is frustrating you?**
As observed in #743, '#' characters should not be allowed as a character to use within an alias. I believe it also doesn't make too much sense to allow other special URL characters such as '?', ',', '=', and ';' either. I created this issue to allow for discussion.

**Your solution**
Create a function that checks if any of these characters are present in custom alias (can perhaps extend `IsValid` method in `CustomAlias` validator). If they are, then throw error.

**Alternatives considered**
Only disallow '#' since it actually causes a problem with long link retrieval and keep allowing the other URL special characters.

**Additional context**
![image](https://user-images.githubusercontent.com/3276350/81229245-7d2e5200-8fbd-11ea-8c08-51209336e304.png)
(note: everything at and after the '#' part of the aliased URL is discarded before sent to server, see #743)

PR #937, opened 2020-07-06
Disallow special URL characters as aliases

arberiii:
### Description
As described on the [issue #744](https://github.com/short-d/short/issues/744), we want to disallow special characters as aliases. For now, the alias is not allowed to contain '#', but the list may be increased in the future.

### Description
Create a map with forbidden characters and this closes [issue #744](https://github.com/short-d/short/issues/744).


backend/app/usecase/validator/alias.go
```diff
@@ -9,6 +9,10 @@ const (
 	customAliasMaxLength = 50
 )
 
+var forbiddenCharacters = map[rune]struct{}{
```
byliuyang, line 12:
```suggestion
var forbiddenCharacters = map[rune]entity.Empty{
```

backend/app/usecase/validator/alias.go
```diff
@@ -9,6 +9,10 @@ const (
...
+	'#': {},
```
byliuyang, line 13:
```suggestion
	'#': entity.Empty{},
```

backend/app/usecase/validator/alias.go
```diff
@@ -28,16 +32,21 @@ func (c CustomAlias) IsValid(alias *string) (bool, Violation) {
...
 		return false, HasFragmentCharacter
 	}
 
 	return true, Valid
 }
 
-// hasFragmentCharacter returns whether the alias contains the '#' character which starts fragment identifiers in URLs
-func (c CustomAlias) hasFragmentCharacter(alias string) bool {
-	return strings.ContainsRune(alias, rune('#'))
+// hasForbiddenCharacter returns whether the alias contains the one of forbidden character which starts fragment identifiers in URLs
```
byliuyang, line 42:
```suggestion
// hasForbiddenCharacter returns whether the alias contains the one of forbidden character
```

backend/app/usecase/validator/alias.go
arberiii: This is redundant.

byliuyang: @alldroll Want it to look better!

PR #938, opened 2020-07-07
Move business logic for short link out of resolver

### LINKED ISSUE #840: [Refactor] Remove business logic from gqlapi resolver
**What is frustrating you?**
The gqlapi resolver uses a singular url input with all fields as optional. Since the create and update functions differ in optional field requirements, the resolver has to format the inputs so that it fits their expected inputs. Thus business logic exists within the resolver which should exist within each use case. 

**Your solution**
Update Creator and Updater to do this business logic instead of the gqlapi resolver.

Coteh:
Fixes #840. 

Also updated the `createUpdate()` method of `ShortLinkInput` to be `GetShortLink()`, and it's now used for both update and create. I did this in order to fix duplicated logic in the create resolver and make it more clear what that method is doing.

Also added `ErrUnknown` as the error type when short link input in either create or update is empty.


backend/app/adapter/gqlapi/resolver/authmutation.go
```diff
@@ -22,51 +21,9 @@ type AuthMutation struct {
...
-		LongLink: s.longLink(),
-		ExpireAt: s.ExpireAt,
-	}
-
-}
-
 // CreateShortLinkArgs represents the possible parameters for CreateShortLink endpoint
 type CreateShortLinkArgs struct {
-	ShortLink ShortLinkInput
+	ShortLink entity.ShortLinkInput
```
byliuyang, line 26: This is part of GraphQL definition. I recommend we define a seperate `ShortLinkInput` inside gql package and make it to `entity. ShortLinkInput` so that when we make changes to Graphql schema in the future, the rest of the system is left untouched.

backend/app/adapter/gqlapi/resolver/authmutation.go
```diff
@@ -77,16 +34,15 @@ func (a AuthMutation) CreateShortLink(args *CreateShortLinkArgs) (*ShortLink, er
...
 
-	longLink := args.ShortLink.longLink()
 	customAlias := args.ShortLink.CustomAlias
-	u := entity.ShortLink{
-		LongLink: longLink,
-		ExpireAt: args.ShortLink.ExpireAt,
+	u := args.ShortLink.GetShortLink()
+	if u == nil {
+		return nil, ErrUnknown{}
 	}
```
byliuyang, line 41: Let's capture the error returned from `CreateShortLink` instead. Ultimately, it is usecase's decision to determine what to do when input is empty.

backend/app/entity/shortlink.go
```diff
@@ -17,3 +17,44 @@ type ShortLink struct {
...
+	if s.IsEmpty() {
+		return nil
+	}
+
+	return &ShortLink{
+		Alias:    s.GetCustomAlias(),
+		LongLink: s.GetLongLink(),
+		ExpireAt: s.ExpireAt,
+	}
+}
```
byliuyang, line 60: I wonder whether this function is actually needed.

backend/app/entity/shortlink.go
```diff
@@ -17,3 +17,44 @@ type ShortLink struct {
...
+
+// ShortLinkInput represents possible ShortLink attributes
+type ShortLinkInput struct {
+	LongLink    *string
+	CustomAlias *string
+	ExpireAt    *time.Time
+}
+
+// IsEmpty checks if ShortLinkInput has no attributes specified.
+func (s *ShortLinkInput) IsEmpty() bool {
```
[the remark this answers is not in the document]
Coteh: You're right. It can definitely be inlined. 👍

backend/app/entity/shortlink.go
Coteh: The logic of getting `ShortLink` from `ShortLinkInput` was being done in two spots (in create and in update). Therefore, I think it makes sense to have this helper function create the `ShortLink` object and remove the duplicate logic.

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: With this change, we now dereference the `ShortLink` object before it's passed into `CreateShortLink`. I throw the `ErrUnknown` error at this point so that we don't get into the problem of dereferencing a null pointer. If we opt for throwing the error from `CreateShortLink` instead, then we should remove the `IsEmpty` method from `ShortLinkInput` and always return a `ShortLink` object from `GetShortLink`.

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: So if I understand correctly, I should just add a simple typedef that points the gql `ShortLInkInput` to the entity one. This is so that in the future, it can be easily swapped out for something else without major refactoring.

backend/app/adapter/gqlapi/resolver/authmutation.go
```diff
@@ -125,9 +81,9 @@ func (a AuthMutation) UpdateShortLink(args *UpdateShortLinkArgs) (*ShortLink, er
 		return nil, ErrInvalidAuthToken{}
 	}
 
-	update := args.ShortLink.createUpdate()
+	update := args.ShortLink.GetShortLink()
 	if update == nil {
-		return nil, nil
+		return nil, ErrUnknown{}
```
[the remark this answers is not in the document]
Coteh: Yes, I think after reading @byliuyang 's suggestion, we should take out the empty check and always return a `ShortLink` update object. Then let `UpdateShortLink` throw an error back to the caller in the case of empty input and let the caller decide what kind of error it is. What do you think?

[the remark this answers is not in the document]
arberiii: I agree!

backend/app/adapter/gqlapi/resolver/authmutation.go
byliuyang: @Coteh Exactly!

backend/app/adapter/gqlapi/resolver/authmutation.go
byliuyang: @Coteh We will have to modify the signature and implementation both `CreateShortLink` and `UpdateShortLink`.

backend/app/entity/shortlink.go
byliuyang: @Coteh I wonder whether we can eliminate the case of getting `ShortLink` from `ShortLinkInput`.

Coteh: I'm not sure because if I understand correctly the reason why we have `ShortLinkInput` is because alias or long link are optional parameters within the GraphQL API. We would have to update the signature for both `CreateShortLink` and `UpdateShortLink` methods to accept a `ShortLinkInput` instead of a `ShortLink` instance, if we want to eliminate getting `ShortLink` from `ShortLinkInput`.

byliuyang: @Coteh I think updating the use cases is a great option. This delegated detailed decisions to the use case layer instead of forcing the API layer to do conversion. Once we converted `input` to non optional representation, the use case won't have the concept of "unset" field any more and won't be able to respond to those scenarios.

byliuyang: I recommend updating the use case if possible.

backend/app/usecase/shortlink/creator.go
```diff
@@ -75,34 +75,33 @@ func (c CreatorPersist) CreateShortLink(shortLink entity.ShortLink, customAlias
 		return entity.ShortLink{}, ErrMaliciousLongLink(longLink)
 	}
 
-	if !c.isAliasProvided(customAlias) {
-		return c.createShortLinkWithAutoAlias(shortLink, user)
+	if !(shortLink.Alias != "") {
+		autoAlias, err := c.createAutoAlias()
+		if err != nil {
+			// TODO create error type for fail create auto alias?
```
Coteh, line 81: @byliuyang Should I create an error type for auto alias fail? Or should I just throw the error up the call stack as-is?

backend/app/usecase/shortlink/creator_test.go
```diff
@@ -226,13 +202,10 @@ func TestShortLinkCreatorPersist_CreateShortLink(t *testing.T) {
 				riskDetector,
 			)
 
-			_, err = shortLinkRepo.GetShortLinkByAlias(testCase.shortLink.Alias)
```
Coteh, line 229: Deleted this short link repo check because it produced a false positive for alias_exists test. Not sure if it's really needed.

Coteh:
@byliuyang Updated the `CreateShortLink` and `UpdateShortLink` interfaces to accommodate for `ShortLinkInput` converting GraphQL input arguments to `ShortLink` instance:
- null strings will be converted to empty strings
- alias will be passed in via `ShortLink` object instead of as a separate argument to `CreateShortLink`

This makes the interface a bit cleaner as a result. Also, I added error handling to `UpdateShortLink` endpoint, similar to what's already been done in `CreateShortLink`. This will provide the extension, error code, and violation with the error message, which it did not have before.

Ready for review.

backend/app/entity/shortlink.go
Coteh: Leaving a comment for myself, @byliuyang clarified over Slack that the use case repository methods for `CreateShortLink` and `UpdateShortLink` should make the decision with short link input, rather than have API layer make the decision and pass it in which is what I have right now. I will resolve this comment when this has been changed.

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: Leaving a comment for myself, @byliuyang clarified over Slack that there should be a new type in `gqlapi` package that represents `shortLinkInput` within GraphQL schema, and instances of this type passed into API layer will be converted to `ShortLinkInput` from `entity` package before being passed into the use case repository methods. This will allow for use case methods to be called from other endpoints such as gRPC in the future.

Coteh: I will resolve this comment once this change has been made.

byliuyang: And allow the APIs to evolve separately.

backend/app/adapter/gqlapi/args/input.go
```diff
@@ -0,0 +1,37 @@
+package args
```
Coteh, line 1: Considering changing name of this package, since it conflicts with resolver function's argument name.

Coteh:
My approach to allowing the GraphQL API and use case to evolve separately is to provide a function in the GraphQL API layer to convert the GraphQL argument type to the entity argument type. The API will call this function to convert its arguments to ones that the entity layer recognizes. The API does not know and does not need to be concerned with what happens with this entity representation.

Ready for review.

backend/app/adapter/gqlapi/args/input.go
byliuyang: input

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -60,6 +61,25 @@ func (e ErrAliasExist) Error() string {
 	return "shortlink alias already exists"
 }
 
+// ErrShortLinkNotFound signifies an expected short link alias does not exist.
+type ErrShortLinkNotFound string
+
+var _ GraphQlError = (*ErrShortLinkNotFound)(nil)
```
byliuyang, line 67:
```suggestion
var _ GraphQLError = (*ErrShortLinkNotFound)(nil)
```

backend/app/entity/shortlink.go
```diff
@@ -17,3 +17,10 @@ type ShortLink struct {
 	OpenGraphTags metatag.OpenGraph
 	TwitterTags   metatag.Twitter
 }
+
+// ShortLinkInput represents possible ShortLink attributes for a new short link or an updated short link.
+type ShortLinkInput struct {
+	LongLink    string
+	CustomAlias string
```
byliuyang, line 24: We need to allow those value to be nullable so that the use case and repository can interpret their meaning themselves. Without nil, there is no clear separation between values not set and set wrong value.

byliuyang: When we update the short link, we will ignore the values if they are nil. If not, we override it with empty string for long link. There is no way to distinguish between those.

byliuyang: What are your thoughts on this?

backend/app/usecase/shortlink/creator.go
```diff
@@ -75,23 +75,24 @@ func (c CreatorPersist) CreateShortLink(createArgs entity.ShortLinkInput, user e
...
-		createArgs.CustomAlias = autoAlias
-	} else {
-		isValid, violation = c.aliasValidator.IsValid(createArgs.CustomAlias)
-		if !isValid {
-			return entity.ShortLink{}, ErrInvalidCustomAlias{createArgs.CustomAlias, violation}
-		}
+		customAlias = autoAlias
 	}
 
 	return c.createShortLink(entity.ShortLink{
```
Coteh, line 93: Adding this comment to remind myself to update `ShortLink` repository methods to accept `ShortLinkInput` instead of `ShortLink`.

backend/app/usecase/shortlink/creator.go
Coteh: This PR is getting pretty big, considering removing this item from scope. Possibly make issue instead?

backend/app/usecase/shortlink/creator_test.go
Coteh: To add a bit more information here: It produced a false positive because existing alias is now passed in alongside the input object, and `GetShortLinkByAlias` checks that.

I think checking for alias existing is a good idea for all other tests, because every other test besides the alias exists test should not have any aliases and we should check that for test correctness.

Coteh: I will add this check back in, but have some sort of way to bypass it only for the `alias_exists` test.

backend/app/entity/shortlink.go
Coteh: As discussed on Slack, we have decided on passing nullable fields into use case, and use case decides what to do with it.

backend/app/adapter/gqlapi/resolver/authmutation.go
```diff
@@ -49,24 +49,24 @@ func (a AuthMutation) CreateShortLink(args *CreateShortLinkArgs) (*ShortLink, er
 		m  shortlink.ErrMaliciousLongLink
 	)
 	if errors.As(err, &ae) {
-		return nil, ErrAliasExist(shortLink.CustomAlias)
+		return nil, ErrAliasExist(ae.Error())
```
byliuyang, line 52: I think we we should avoid leaking the exact error to the API consumer to prevent hackers from launching cyber attack. Can we change back to what we had before?

backend/app/adapter/gqlapi/resolver/authmutation.go
```diff
@@ -91,19 +91,19 @@ func (a AuthMutation) UpdateShortLink(args *UpdateShortLinkArgs) (*ShortLink, er
 		nf shortlink.ErrShortLinkNotFound
 	)
 	if errors.As(err, &ae) {
-		return nil, ErrAliasExist(update.CustomAlias)
+		return nil, ErrAliasExist(ae.Error())
 	}
 	if errors.As(err, &l) {
-		return nil, ErrInvalidLongLink{update.LongLink, string(l.Violation)}
+		return nil, ErrInvalidLongLink{l.Error(), string(l.Violation)}
```
byliuyang, line 97: Ditto

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -8,6 +8,7 @@ const (
 	ErrCodeUnknown            ErrCode = "unknown"
 	ErrCodeAliasAlreadyExist          = "aliasAlreadyExist"
 	ErrCodeShortLinkNotFound          = "shortLinkNotFound"
+	ErrCodeNewAliasNotSpecified		  = "newAliasNotSpecified"
```
byliuyang: Format

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -80,10 +81,28 @@ func (e ErrShortLinkNotFound) Error() string {
 	return "shortlink does not exist"
 }
 
+// ErrNewAliasNotSpecified signifies that user provided an empty alias when updating short link
+type ErrNewAliasNotSpecified string
```
byliuyang, line 85:
```suggestion
type ErrEmptyAlias string
```

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -80,10 +81,28 @@ func (e ErrShortLinkNotFound) Error() string {
 	return "shortlink does not exist"
 }
 
+// ErrNewAliasNotSpecified signifies that user provided an empty alias when updating short link
```
byliuyang, line 84: Let's not to tie this error to specific use cases.

backend/app/usecase/shortlink/creator_test.go
```diff
@@ -21,11 +21,23 @@ func TestShortLinkCreatorPersist_CreateShortLink(t *testing.T) {
...
+	// Set of aliases representing different classes of aliases to be tested.
 	alias := "220uFicCJj"
 	longAlias := "an-alias-cannot-be-used-to-specify-default-arguments"
 	invalidFragmentAlias := "cant-have#chr"
 	emptyAlias := ""
 
+	// Set of long links representing different classes of long links to be tested.
+	longLink := "https://www.google.com"
+	invalidLongLink := "aaaaaaaaaaaaaaaaaaa"
+	maliciousLongLink := "http://malware.wicar.org/data/ms14_064_ole_not_xp.html"
```
byliuyang, line 39: Nit: We can create helper to convert string literal to pointer in the future to remove this tech debt.

backend/app/usecase/shortlink/creator_test.go
```diff
@@ -161,12 +192,12 @@ func TestShortLinkCreatorPersist_CreateShortLink(t *testing.T) {
...
 			shortLinkArgs: entity.ShortLinkInput{
-				CustomAlias: "220uFicCJj",
-				LongLink:    "http://malware.wicar.org/data/ms14_064_ole_not_xp.html",
+				CustomAlias: &alias,
+				LongLink:    &maliciousLongLink,
 				ExpireAt:    &now,
 			},
 			blockedLongLinks: map[string]bool{
-				"http://malware.wicar.org/data/ms14_064_ole_not_xp.html": true,
+				maliciousLongLink: true,
```
byliuyang, line 200: It seems that this can be embedded in the test case directly.

backend/app/usecase/shortlink/updater.go
```diff
@@ -46,21 +53,29 @@ func (u UpdaterPersist) UpdateShortLink(
 		return entity.ShortLink{}, ErrShortLinkNotFound("short link not found")
 	}
 
-	aliasExist, err := u.shortLinkRepo.IsAliasExist(updateArgs.CustomAlias)
-	if err != nil {
-		return entity.ShortLink{}, err
+	newAlias := updateArgs.GetCustomAlias(oldAlias)
+	if newAlias == "" {
+		return entity.ShortLink{}, ErrNewAliasNotSpecified("new alias not specified")
```
byliuyang, line 58:
```suggestion
		return entity.ShortLink{}, ErrEmptyAlias("alias is empty")
```

backend/app/usecase/shortlink/updater.go
```diff
@@ -46,21 +53,29 @@ func (u UpdaterPersist) UpdateShortLink(
...
+
+	// Only check if it exists if user is changing the alias to something else
+	if (newAlias != oldAlias) {
+		aliasExist, err := u.shortLinkRepo.IsAliasExist(newAlias)
+		if err != nil {
+			return entity.ShortLink{}, err
+		}
+		if aliasExist {
+			return entity.ShortLink{}, ErrAliasExist("short link alias already exists")
+		}
```
byliuyang: This check can be removed now. We are already checking the existence of short link ownership. Our DB ensures short link exists when mapping for that short link exists.

backend/app/adapter/gqlapi/input/input.go
```diff
@@ -0,0 +1,23 @@
+package input
```
byliuyang, line 1: Let's rename this file to shortlink.

backend/app/adapter/gqlapi/resolver/error.go
```diff
@@ -5,19 +5,21 @@ type ErrCode string
...
-	ErrCodeUnknown            ErrCode = "unknown"
-	ErrCodeAliasAlreadyExist          = "aliasAlreadyExist"
-	ErrCodeRequesterNotHuman          = "requesterNotHuman"
-	ErrCodeInvalidLongLink            = "invalidLongLink"
-	ErrCodeInvalidCustomAlias         = "invalidCustomAlias"
-	ErrCodeAliasWithFragment          = "aliasWithFragment"
-	ErrCodeMaliciousContent           = "maliciousContent"
-	ErrCodeInvalidAuthToken           = "invalidAuthToken"
-	ErrCodeUnauthorizedAction         = "unauthorizedAction"
+	ErrCodeUnknown              ErrCode = "unknown"
```
byliuyang, line 8: Let's format the code

backend/app/usecase/shortlink/creator.go
```diff
@@ -64,9 +64,9 @@ type CreatorPersist struct {
 
 // CreateShortLink persists a new short link with a given or auto generated alias in the repository.
 // TODO(issue#235): add functionality for public URLs
-func (c CreatorPersist) CreateShortLink(shortLink entity.ShortLink, customAlias *string, user entity.User, isPublic bool) (entity.ShortLink, error) {
-	longLink := shortLink.LongLink
-	isValid, violation := c.longLinkValidator.IsValid(&longLink)
+func (c CreatorPersist) CreateShortLink(createArgs entity.ShortLinkInput, user entity.User, isPublic bool) (entity.ShortLink, error) {
```
byliuyang, line 67:
```suggestion
func (c CreatorPersist) CreateShortLink(shortLinkInput entity.ShortLinkInput, user entity.User, isPublic bool) (entity.ShortLink, error) {
```

backend/app/usecase/shortlink/creator.go
```diff
@@ -47,7 +47,7 @@ func (e ErrMaliciousLongLink) Error() string {
 
 // Creator represents a ShortLink alias creator
 type Creator interface {
-	CreateShortLink(shortLink entity.ShortLink, alias *string, user entity.User, isPublic bool) (entity.ShortLink, error)
+	CreateShortLink(createArgs entity.ShortLinkInput, user entity.User, isPublic bool) (entity.ShortLink, error)
```
byliuyang, line 50:
```suggestion
	CreateShortLink(shortLinkInput entity.ShortLinkInput, user entity.User, isPublic bool) (entity.ShortLink, error)
```

backend/app/usecase/shortlink/creator.go
```diff
@@ -75,34 +75,38 @@ func (c CreatorPersist) CreateShortLink(shortLink entity.ShortLink, customAlias
...
+
+	return c.createShortLink(entity.ShortLink{
+		LongLink: longLink,
+		Alias:    customAlias,
+		ExpireAt: createArgs.ExpireAt,
+	}, user)
 }
 
-func (c CreatorPersist) createShortLinkWithAutoAlias(shortLink entity.ShortLink, user entity.User) (entity.ShortLink, error) {
+func (c CreatorPersist) createAutoAlias() (string, error) {
```
byliuyang, line 100:
```suggestion
func (c CreatorPersist) generateAlias() (string, error) {
```

backend/app/usecase/shortlink/updater.go
```diff
@@ -13,11 +11,22 @@ import (
...
+type ErrNewAliasNotSpecified string
+
+func (e ErrNewAliasNotSpecified) Error() string {
+	return string(e)
+}
 
 // Updater mutates existing short links.
 type Updater interface {
-	UpdateShortLink(oldAlias string, update entity.ShortLink, user entity.User) (entity.ShortLink, error)
+	UpdateShortLink(oldAlias string, updateArgs entity.ShortLinkInput, user entity.User) (entity.ShortLink, error)
```
byliuyang, line 29:
```suggestion
	UpdateShortLink(oldAlias string, shortLinkInput entity.ShortLinkInput, user entity.User) (entity.ShortLink, error)
```

backend/app/usecase/shortlink/updater.go
```diff
@@ -33,39 +42,47 @@ type UpdaterPersist struct {
 // UpdateShortLink mutates a short link in the repository.
 func (u UpdaterPersist) UpdateShortLink(
 	oldAlias string,
-	update entity.ShortLink,
+	updateArgs entity.ShortLinkInput,
```
byliuyang, line 45:
```suggestion
	shortLinkInput entity.ShortLinkInput,
```

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -19,12 +19,27 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
 
 	now := time.Now().UTC()
 
+	// ShortLink creator and updater methods require string pointers
+	// It is not possible to reference a literal string directly in Go
+	// To make the test suite as clean as possible, they are instead declared below and
+	// the variables are passed into the test case by reference, which is a legal semantic in Go.
+	// https://golang.org/ref/spec#Address_operators
```
byliuyang, line 26: Let's define 
```
func String(str string) *string {
	return &str
}
```
inside `backend/app/fw/ptr/string.go` and use this function to directly embed string in the test case.

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: I did this to prevent panic from dereferencing null pointer. I'll change it back and add some conditional logic to prevent null dereference.

backend/app/adapter/gqlapi/resolver/error.go
Coteh: I believe I ran script formatter already, but I'll rerun it after I make my changes.

backend/app/adapter/gqlapi/resolver/authmutation.go
Coteh: Adding another comment to mention that no need to add more conditional logic, I reused the `GetLongLink` and `GetCustomAlias` methods with "" as `defaultVal`.

Coteh: Will push this out along with my other changes in a bit.

backend/app/usecase/shortlink/creator.go
Coteh: Issue created. https://github.com/short-d/short/issues/950

Will update the TODO comment with issue number.

backend/app/adapter/sqldb/user_short_link.go
```diff
@@ -29,7 +29,7 @@ VALUES ($1,$2)
 		table.UserShortLink.ColumnShortLinkAlias,
 	)
 
-	_, err := u.db.Exec(statement, user.ID, shortLink.Alias)
+	_, err := u.db.Exec(statement, user.ID, shortLinkInput.GetCustomAlias(""))
```
Coteh, line 32: I wonder if I should instead pass the pointer `shortLinkInput.CustomAlias` directly here. It should never end up being `null` but if for some reason it is then the relationship should reflect that... 🤔

backend/app/usecase/shortlink/updater.go
Coteh: This check is needed because we need to check if the new alias exists. If this check is removed then unknown err will be returned from the API is user attempts to change alias to an alias that already exists.

backend/app/adapter/gqlapi/input/shortlink.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// ShortLinkInput represents possible ShortLink attributes
+type ShortLinkInput struct {
+	LongLink    *string
+	CustomAlias *string
+	ExpireAt    *time.Time
+}
+
+// CreateShortLinkInput will massage the GraphQL arguments into a form that use case will accept.
```
byliuyang, line 16:
```suggestion
// CreateShortLinkInput converts GraphQL ShortLinkInput into consumable entity for use cases.
```

backend/app/adapter/sqldb/short_link.go
```diff
@@ -94,42 +94,28 @@ WHERE "%s"=$1;`,
...
-		shortLink.ExpireAt,
-		shortLink.CreatedAt,
-		shortLink.UpdatedAt,
-		shortLink.OpenGraphTags.Title,
-		shortLink.OpenGraphTags.Description,
-		shortLink.OpenGraphTags.ImageURL,
-		shortLink.TwitterTags.Title,
-		shortLink.TwitterTags.Description,
-		shortLink.TwitterTags.ImageURL,
+		shortLinkInput.CustomAlias,
```
byliuyang, line 109:
```suggestion
		shortLinkInput.GetCustomAlias(""),
```
I think this can't be nil.

backend/app/adapter/sqldb/short_link.go
```diff
@@ -94,42 +94,28 @@ WHERE "%s"=$1;`,
...
+		shortLinkInput.LongLink,
```
byliuyang, line 110:
```suggestion
		shortLinkInput.GetLongLink(""),
```

backend/app/adapter/sqldb/short_link.go
```diff
@@ -144,18 +130,23 @@ WHERE "%s"=$5;`,
 
 	_, err := s.db.Exec(
 		statement,
-		newShortLink.Alias,
-		newShortLink.LongLink,
-		newShortLink.ExpireAt,
-		newShortLink.UpdatedAt,
+		shortLinkInput.CustomAlias,
```
byliuyang, line 133:
```suggestion
		shortLinkInput.GetCustomAlias(""),
```

backend/app/adapter/sqldb/short_link.go
```diff
@@ -144,18 +130,23 @@ WHERE "%s"=$5;`,
...
+		shortLinkInput.LongLink,
```
byliuyang, line 134:
```suggestion
		shortLinkInput.GetLongLink(""),
```

backend/app/adapter/sqldb/short_link.go
```diff
@@ -144,18 +130,23 @@ WHERE "%s"=$5;`,
...
 		oldAlias,
 	)
 
 	if err != nil {
 		return entity.ShortLink{}, err
 	}
 
-	return newShortLink, nil
+	return entity.ShortLink{
+		Alias:     *shortLinkInput.CustomAlias,
```
byliuyang, line 145:
```suggestion
		Alias:     shortLinkInput.GetCustomAlias(""),
```

backend/app/adapter/sqldb/short_link.go
```diff
@@ -144,18 +130,23 @@ WHERE "%s"=$5;`,
...
+		LongLink:  *shortLinkInput.LongLink,
```
byliuyang, line 146:
```suggestion
		LongLink:  shortLinkInput.GetLongLink(""),
```

backend/app/adapter/sqldb/short_link_integration_test.go
```diff
@@ -540,9 +520,12 @@ func TestShortLinkSql_CreateShortLink(t *testing.T) {
 
 					assert.Equal(t, nil, err)
 
-					shortLink, err := shortLinkRepo.GetShortLinkByAlias(testCase.shortLink.Alias)
+					shortLink, err := shortLinkRepo.GetShortLinkByAlias(*testCase.shortLinkInput.CustomAlias)
```
byliuyang, line 523:
```suggestion
					shortLink, err := shortLinkRepo.GetShortLinkByAlias(testCase.shortLinkInput.GetCustomAlias(""))
```

backend/app/usecase/shortlink/updater.go
Coteh: Also, `alias_already_exists` test in `updater_test.go` fails if this check is removed.

backend/app/adapter/sqldb/short_link_integration_test.go
```diff
@@ -444,15 +445,12 @@ func TestShortLinkSql_CreateShortLink(t *testing.T) {
 	title1 := "title1"
 	description1 := "description1"
 	imageURL1 := "url1"
```
byliuyang, line 447: We can also inline those with `ptr.String("")`

backend/app/adapter/sqldb/short_link_integration_test.go
```diff
@@ -630,11 +613,11 @@ func TestShortLinkSql_UpdateShortLink(t *testing.T) {
 					shortLinkRepo := sqldb.NewShortLinkSQL(sqlDB)
 					shortLink, err := shortLinkRepo.UpdateShortLink(
 						testCase.oldAlias,
-						testCase.newShortLink,
+						testCase.shortLinkInput,
 					)
 					assert.Equal(t, nil, err)
 
-					shortLink, err = shortLinkRepo.GetShortLinkByAlias(testCase.newShortLink.Alias)
+					shortLink, err = shortLinkRepo.GetShortLinkByAlias(*testCase.shortLinkInput.CustomAlias)
```
byliuyang, line 620:
```suggestion
					shortLink, err = shortLinkRepo.GetShortLinkByAlias(testCase.shortLinkInput.GetCustomAlias(""))
```

backend/app/adapter/sqldb/user_short_link.go
byliuyang: I think it's safe and make sense to use `GetCustomAlias`. It's doesn't accept `nil`.

backend/app/entity/shortlink.go
```diff
@@ -17,3 +17,28 @@ type ShortLink struct {
 	OpenGraphTags metatag.OpenGraph
 	TwitterTags   metatag.Twitter
 }
+
+// ShortLinkInput represents possible ShortLink attributes for a new short link or an updated short link.
```
byliuyang, line 21:
```suggestion
// ShortLinkInput represents possible ShortLink attributes a short link.
```
We don't mention how we are going to use it here. That's the job and freedom of use case.

backend/app/entity/shortlink.go
```diff
@@ -17,3 +17,28 @@ type ShortLink struct {
...
+type ShortLinkInput struct {
+	LongLink    *string
+	CustomAlias *string
+	ExpireAt    *time.Time
+	CreatedAt   *time.Time
+	UpdatedAt   *time.Time
+}
+
+// GetLongLink fetches LongLink for ShortLinkInput, with empty string as the default value.
```
byliuyang, line 30:
```suggestion
// GetLongLink fetches LongLink for ShortLinkInput with default value.
```
We are passing `defaultVal` here, which not necessarily empty string.

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -21,15 +21,23 @@ func (s ShortLinkFake) IsAliasExist(alias string) (bool, error) {
...
+	}
+	isExist, err := s.IsAliasExist(*shortLinkInput.CustomAlias)
 	if err != nil {
 		return err
 	}
 	if isExist {
 		return errors.New("alias exists")
 	}
-	s.shortLinks[shortLink.Alias] = shortLink
+	s.shortLinks[*shortLinkInput.CustomAlias] = entity.ShortLink{
```
byliuyang, line 35:
```suggestion
        customAlias := shortLinkInput.GetCustomAlias("")
	s.shortLinks[customAlias] = entity.ShortLink{
```

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -21,15 +21,23 @@ func (s ShortLinkFake) IsAliasExist(alias string) (bool, error) {
...
+		Alias:     *shortLinkInput.CustomAlias,
```
byliuyang, line 36:
```suggestion
		Alias:     customAlias,
```

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -149,6 +135,96 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
...
+				{
+					Alias:     "git",
+					LongLink:  "https://github.com/short-d",
+					UpdatedAt: &now,
+				},
+			},
+			expectedHasErr: true,
+		},
+		{
+			name:  "long link not changed if not specified",
```
Coteh, line 168: This test currently fails when it should pass. Leaving a review comment for myself to investigate why the user relation check on line 396 of `updater_test.go` fails when this short link is updated.

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -75,9 +87,9 @@ func (s ShortLinkFake) UpdateShortLink(oldAlias string, newShortLink entity.Shor
 	createdBy := prevShortLink.CreatedBy
 	createdAt := prevShortLink.CreatedAt
 	return entity.ShortLink{
-		Alias:     newShortLink.Alias,
-		LongLink:  newShortLink.LongLink,
-		ExpireAt:  newShortLink.ExpireAt,
+		Alias:     *shortLinkInput.CustomAlias,
```
byliuyang, line 90:
```suggestion
		Alias:     shortLinkInput.GetCustomAlias(""),
```

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -75,9 +87,9 @@ func (s ShortLinkFake) UpdateShortLink(oldAlias string, newShortLink entity.Shor
...
+		LongLink:  *shortLinkInput.LongLink,
```
byliuyang, line 91:
```suggestion
		LongLink:  shortLinkInput.GetLongLink(""),
```

backend/app/usecase/repository/usershortlink_fake.go
```diff
@@ -15,12 +15,24 @@ type UserShortLinkFake struct {
...
+	if err != nil {
+		return err
+	}
+	if isExist {
 		return errors.New("relationship exists")
 	}
 	u.users = append(u.users, user)
-	u.shortLinks = append(u.shortLinks, shortLink)
+	u.shortLinks = append(u.shortLinks, entity.ShortLink{
+		Alias:     *shortLinkInput.CustomAlias,
```
byliuyang, line 31:
```suggestion
		Alias:     shortLinkInput.GetCustomAlias(""),
```

backend/app/usecase/shortlink/creator.go
```diff
@@ -64,9 +64,9 @@ type CreatorPersist struct {
 
 // CreateShortLink persists a new short link with a given or auto generated alias in the repository.
 // TODO(issue#235): add functionality for public URLs
-func (c CreatorPersist) CreateShortLink(shortLink entity.ShortLink, customAlias *string, user entity.User, isPublic bool) (entity.ShortLink, error) {
-	longLink := shortLink.LongLink
-	isValid, violation := c.longLinkValidator.IsValid(&longLink)
+func (c CreatorPersist) CreateShortLink(shortLinkInput entity.ShortLinkInput, user entity.User, isPublic bool) (entity.ShortLink, error) {
+	longLink := shortLinkInput.GetLongLink("")
+	isValid, violation := c.longLinkValidator.IsValid(longLink)
```
byliuyang, line 69: `IsValid` is accepting a pointer. We should pass in the pointer directly.

backend/app/usecase/shortlink/creator.go
```diff
@@ -75,34 +75,37 @@ func (c CreatorPersist) CreateShortLink(shortLink entity.ShortLink, customAlias
 		return entity.ShortLink{}, ErrMaliciousLongLink(longLink)
 	}
 
-	if !c.isAliasProvided(customAlias) {
-		return c.createShortLinkWithAutoAlias(shortLink, user)
-	}
-
+	customAlias := shortLinkInput.GetCustomAlias("")
 	isValid, violation = c.aliasValidator.IsValid(customAlias)
```
byliuyang, line 79: IsValid is accepting a pointer. We should pass in the pointer directly.

backend/app/usecase/shortlink/creator.go
```diff
@@ -75,34 +75,37 @@ func (c CreatorPersist) CreateShortLink(shortLink entity.ShortLink, customAlias
...
-	return c.createShortLinkWithCustomAlias(shortLink, randomAlias, user)
+	return string(key), nil
 }
 
-func (c CreatorPersist) createShortLinkWithCustomAlias(shortLink entity.ShortLink, alias string, user entity.User) (entity.ShortLink, error) {
-	shortLink.Alias = alias
-
-	isExist, err := c.shortLinkRepo.IsAliasExist(alias)
+func (c CreatorPersist) createShortLink(shortLinkInput entity.ShortLinkInput, user entity.User) (entity.ShortLink, error) {
+	isExist, err := c.shortLinkRepo.IsAliasExist(*shortLinkInput.CustomAlias)
```
byliuyang, line 108:
```suggestion
	isExist, err := c.shortLinkRepo.IsAliasExist(shortLinkInput.GetCustomAlias(""))
```

backend/app/usecase/shortlink/creator.go
```diff
@@ -112,15 +115,20 @@ func (c CreatorPersist) createShortLinkWithCustomAlias(shortLink entity.ShortLin
...
+	err = c.shortLinkRepo.CreateShortLink(shortLinkInput)
 	if err != nil {
 		return entity.ShortLink{}, err
 	}
 
-	err = c.userShortLinkRepo.CreateRelation(user, shortLink)
-	return shortLink, err
+	err = c.userShortLinkRepo.CreateRelation(user, shortLinkInput)
+	return entity.ShortLink{
+		LongLink:  *shortLinkInput.LongLink,
```
byliuyang, line 127:
```suggestion
		LongLink:  shortLinkInput.GetLongLink(""),
```

backend/app/usecase/shortlink/creator.go
```diff
@@ -112,15 +115,20 @@ func (c CreatorPersist) createShortLinkWithCustomAlias(shortLink entity.ShortLin
...
+		Alias:     *shortLinkInput.CustomAlias,
```
byliuyang, line 128:
```suggestion
		Alias:     shortLinkInput.GetCustomAlias(""),
```

backend/app/usecase/shortlink/updater.go
```diff
@@ -33,39 +42,47 @@ type UpdaterPersist struct {
...
 
 	shortLink, err := u.shortLinkRepo.GetShortLinkByAlias(oldAlias)
 	if err != nil {
 		return entity.ShortLink{}, err
 	}
 
-	shortLink = u.updateAlias(shortLink, update)
-	shortLink = u.updateLongLink(shortLink, update)
+	shortLink.Alias = newAlias
+	shortLink.LongLink = shortLinkInput.GetLongLink(shortLink.LongLink)
```
byliuyang, line 78:
```suggestion
	longLink = shortLinkInput.GetLongLink(shortLink.LongLink)
```

backend/app/usecase/shortlink/updater.go
```diff
@@ -33,39 +42,47 @@ type UpdaterPersist struct {
...
 
-	isValid, violation := u.aliasValidator.IsValid(&shortLink.Alias)
+	isValid, violation := u.aliasValidator.IsValid(shortLink.Alias)
```
byliuyang, line 80:
```suggestion
	isValid, violation := u.aliasValidator.IsValid(newAlias)
```

backend/app/usecase/shortlink/updater.go
```diff
@@ -33,39 +42,47 @@ type UpdaterPersist struct {
...
 	if !isValid {
 		return entity.ShortLink{}, ErrInvalidCustomAlias{shortLink.Alias, violation}
 	}
 
-	isValid, violation = u.longLinkValidator.IsValid(&shortLink.LongLink)
+	isValid, violation = u.longLinkValidator.IsValid(shortLink.LongLink)
```
byliuyang, line 85:
```suggestion
	isValid, violation = u.longLinkValidator.IsValid(longLink)
```

backend/app/usecase/shortlink/updater.go
```diff
@@ -75,25 +92,13 @@ func (u UpdaterPersist) UpdateShortLink(
...
-}
-
-func (u *UpdaterPersist) updateLongLink(shortLink, update entity.ShortLink) entity.ShortLink {
-	newLongLink := update.LongLink
-	if newLongLink != "" {
-		shortLink.LongLink = newLongLink
-	}
-	return shortLink
+	return u.shortLinkRepo.UpdateShortLink(oldAlias, entity.ShortLinkInput{
+		CustomAlias: &shortLink.Alias,
```
byliuyang, line 97:
```suggestion
		CustomAlias: &newAlias,
```

backend/app/usecase/shortlink/updater.go
```diff
@@ -75,25 +92,13 @@ func (u UpdaterPersist) UpdateShortLink(
...
+		LongLink:    &shortLink.LongLink,
```
byliuyang, line 98:
```suggestion
		LongLink:    &longLink,
```

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -19,12 +19,27 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
...
+	// https://golang.org/ref/spec#Address_operators
+
+	// Set of aliases representing different classes of aliases to be tested.
+	alias := "short-d"
+	fragmentAlias := "#http-bin"
+
+	// Set of long links representing different classes of long links to be tested.
+	longLink := "https://httpbin.org/get?p1=v1"
+	invalidLongLink := "aaaaaaaaaaaaaaaaaaa"
+	maliciousLongLink := "http://malware.wicar.org/data/ms14_064_ole_not_xp.html"
```
byliuyang: Let's replace this with `ptr.String`.

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -19,12 +19,27 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
...
+
 	testCases := []struct {
 		name               string
 		alias              string
 		shortlinks         shortLinks
 		user               entity.User
-		update             entity.ShortLink
+		updateArgs         entity.ShortLinkInput
```
byliuyang, line 28:
```suggestion
		shortLinkInput        entity.ShortLinkInput
```

backend/app/usecase/shortlink/updater.go
byliuyang: @Coteh You are right. I misread this!

backend/app/adapter/sqldb/short_link.go
Coteh: 👍 I think you're right. The column is non-nullable in both `short_link` and `user_short_link` tables.

backend/app/adapter/sqldb/short_link.go
Coteh: Yes this is needed, thanks for pointing this out. Although long link is nullable in SQL, application will run into an error when converting the null value to string.

backend/app/adapter/sqldb/user_short_link.go
Coteh: After reading your other comments and reviewing the structure of the SQL tables, you're right. I will mark this as resolved.

backend/app/entity/shortlink.go
Coteh: Nice catch. I wrote that docstring before I changed it. Thanks! 👍

backend/app/usecase/shortlink/creator.go
Coteh: I updated validator interface to accept value instead of pointer, therefore it's ok to pass in value directly. What do you think? Should we change interface back to accepting pointers to check for `nil`?

backend/app/usecase/shortlink/creator_test.go
Coteh: @byliuyang Just wanted to mark this as unresolved briefly. So do you think `shortLinkRepo.GetShortLinkByAlias` on line 229 should be added back in? Or is the `userShortLinkRepo.HasMapping` check below it sufficient for checking short link existence?

backend/app/entity/shortlink.go
```diff
@@ -17,3 +17,28 @@ type ShortLink struct {
...
+
+// GetLongLink fetches LongLink for ShortLinkInput with default value.
+func (s *ShortLinkInput) GetLongLink(defaultVal string) string {
+	if s.LongLink == nil {
+		return defaultVal
+	}
+	return *s.LongLink
+}
+
+// GetCustomAlias fetches CustomAlias for ShortLinkInput, with empty string as the default value.
```
byliuyang, line 38:
```suggestion
// GetCustomAlias fetches CustomAlias for ShortLinkInput with default value.
```

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -21,15 +21,24 @@ func (s ShortLinkFake) IsAliasExist(alias string) (bool, error) {
...
+	}
+	isExist, err := s.IsAliasExist(*shortLinkInput.CustomAlias)
 	if err != nil {
 		return err
 	}
 	if isExist {
 		return errors.New("alias exists")
 	}
-	s.shortLinks[shortLink.Alias] = shortLink
+        customAlias := shortLinkInput.GetCustomAlias("")
```
byliuyang, line 35:
```suggestion
       customAlias := shortLinkInput.GetCustomAlias("")
```

backend/app/usecase/repository/usershortlink_fake.go
```diff
@@ -15,12 +15,24 @@ type UserShortLinkFake struct {
 }
 
 // CreateRelation creates many to many relationship between User and ShortLink.
-func (u *UserShortLinkFake) CreateRelation(user entity.User, shortLink entity.ShortLink) error {
-	if u.IsRelationExist(user, shortLink) {
+func (u *UserShortLinkFake) CreateRelation(user entity.User, shortLinkInput entity.ShortLinkInput) error {
+	if shortLinkInput.CustomAlias == nil {
+		return errors.New("empty alias")
+	}
+	isExist, err := u.HasMapping(user, *shortLinkInput.CustomAlias)
```
byliuyang, line 22:
```suggestion
	isExist, err := u.HasMapping(user, shortLinkInput.GetCustomAlias(""))
```

backend/app/usecase/shortlink/creator.go
byliuyang: @Coteh Then it's good to go. Let's check whether the alias is `nil` and generate one at the beginning of this function. We generate when alias is not provided.

byliuyang: I mean line 84 - line 91

backend/app/usecase/shortlink/updater.go
```diff
@@ -33,39 +42,46 @@ type UpdaterPersist struct {
...
+	shortLinkInput entity.ShortLinkInput,
 	user entity.User,
 ) (entity.ShortLink, error) {
 	hasMapping, err := u.userShortLinkRepo.HasMapping(user, oldAlias)
 	if err != nil {
 		return entity.ShortLink{}, err
 	}
 	if !hasMapping {
-		return entity.ShortLink{}, ErrShortLinkNotFound
+		return entity.ShortLink{}, ErrShortLinkNotFound("short link not found")
```
byliuyang, line 53: We need to include the alias so that we can debug when something goes wrong. `ErrShortLinkNotFound` already communicated the error type.

backend/app/usecase/repository/shortlink_fake.go
Coteh: Not sure why this was spaced out and wasn't caught by the formatter. Thanks for catching this!

backend/app/usecase/shortlink/creator.go
Coteh: Right now, if alias is `nil`, then `customAlias` will be empty string, and if custom alias is empty string then alias will be auto-generated. If I change the logic to what you suggested, then empty string will be a valid alias. Is this something you want?

backend/app/usecase/shortlink/creator.go
Coteh: Didn't see this comment earlier. As mentioned in the other comment, `validator` now accepts value rather than pointer. Should I change the interface back to accepting pointer?

backend/app/usecase/shortlink/updater_test.go
Coteh: After some investigation, this is failing because the fake short link repository does not update the user_short_link relation table. The SQL version implicitly updates the user_short_link relation table via `ON UPDATE CASCADE` trigger. I will think of a way to implement this cleanly using the current fake repo implementation, or I'll make a new issue and move this test there.

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -238,8 +329,8 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
 				ID:    "1",
 				Email: "gopher@golang.org",
 			},
-			update: entity.ShortLink{
-				LongLink: "http://malware.wicar.org/data/ms14_064_ole_not_xp.html",
+			shortLinkInput: entity.ShortLinkInput{
```
Coteh, line 332: "reject malicious long link" now fails. There seems to be no error coming out of it when it should. I will investigate and resolve this comment when it's fixed.

Coteh: Fixed. It was using the old long link reference.

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -21,15 +21,24 @@ func (s ShortLinkFake) IsAliasExist(alias string) (bool, error) {
 }
 
 // CreateShortLink inserts a new ShortLink into short_link table.
-func (s *ShortLinkFake) CreateShortLink(shortLink entity.ShortLink) error {
-	isExist, err := s.IsAliasExist(shortLink.Alias)
+func (s *ShortLinkFake) CreateShortLink(shortLinkInput entity.ShortLinkInput) error {
+	if shortLinkInput.CustomAlias == nil {
+		return errors.New("alias empty")
+	}
+	isExist, err := s.IsAliasExist(*shortLinkInput.CustomAlias)
```
byliuyang, line 28:
```suggestion
	isExist, err := s.IsAliasExist(customAlias)
```
Let's move line 35 before this one.

backend/app/usecase/repository/usershortlink_fake.go
```diff
@@ -15,12 +15,24 @@ type UserShortLinkFake struct {
 }
 
 // CreateRelation creates many to many relationship between User and ShortLink.
-func (u *UserShortLinkFake) CreateRelation(user entity.User, shortLink entity.ShortLink) error {
-	if u.IsRelationExist(user, shortLink) {
+func (u *UserShortLinkFake) CreateRelation(user entity.User, shortLinkInput entity.ShortLinkInput) error {
+	if shortLinkInput.CustomAlias == nil {
+		return errors.New("empty alias")
+	}
+	isExist, err := u.HasMapping(user, shortLinkInput.GetCustomAlias(""))
```
byliuyang, line 22:
```suggestion
        customAlias := shortLinkInput.GetCustomAlias("")
	isExist, err := u.HasMapping(user, customAlias)
```

backend/app/usecase/repository/usershortlink_fake.go
```diff
@@ -15,12 +15,24 @@ type UserShortLinkFake struct {
...
+	if err != nil {
+		return err
+	}
+	if isExist {
 		return errors.New("relationship exists")
 	}
 	u.users = append(u.users, user)
-	u.shortLinks = append(u.shortLinks, shortLink)
+	u.shortLinks = append(u.shortLinks, entity.ShortLink{
+		Alias:     shortLinkInput.GetCustomAlias(""),
```
byliuyang, line 31:
```suggestion
		Alias:     customAlias,
```

backend/app/usecase/shortlink/updater_test.go
byliuyang: Let's pass in fake short link repo as a dependency of `user_short_link` and add a helper method to update the alias for now. And add a TODO to switch to eventbus for change propagation across multiple repositories.

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -12,6 +12,14 @@ var _ ShortLink = (*ShortLinkFake)(nil)
 // ShortLinkFake accesses ShortLink information in short_link table through SQL.
 type ShortLinkFake struct {
 	shortLinks map[string]entity.ShortLink
+	// TODO(issue#958) use eventbus for propagating short link change to user short link repo
```
byliuyang, line 15:
```suggestion
	// TODO(issue#958) use eventbus for propagating short link change to all related repos
```

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -12,6 +12,14 @@ var _ ShortLink = (*ShortLinkFake)(nil)
...
+	userShortLinkRepoFake *UserShortLinkFake
+}
+
+// ProvideUserShortLinkRepoFake allows ShortLinkFake to propagate new alias to user short link repo.
```
byliuyang, line 19: It's a dependency for `ShortLinkFake`. Can we inject it through the constructor? If you just trying to fix this test. I am okey with keeping what you have for now.

backend/app/usecase/repository/usershortlink_fake.go
```diff
@@ -71,7 +72,7 @@ func (u UserShortLinkFake) UpdateRelation(user entity.User, oldAlias string, sho
 			return nil
 		}
 	}
-	return errors.New("relationship does not exist")
+	return errors.New("no relationships with this alias exist")
```
byliuyang, line 75: We should pass in the actually alias in the error.

backend/app/usecase/repository/shortlink_fake.go
Coteh: Right, I can just set it inside `NewShortLinkFake`. 👍

backend/app/usecase/repository/shortlink_fake.go
Coteh: Yes that makes sense, as user short link is not the only thing connected to short link repo, like we discussed yesterday. Thanks for this suggestion!

backend/app/usecase/shortlink/creator_test.go
Coteh: Resolved this by adding a `shouldAliasExist` bool in test case. Only `alias_exists` test will have this set as true since alias is expected to be there for that test. All other tests should ensure that alias does not already exist, as a precondition check.

backend/app/usecase/repository/shortlink_fake.go
```diff
@@ -65,28 +76,39 @@ func (s ShortLinkFake) GetShortLinksByAliases(aliases []string) ([]entity.ShortL
...
+		ExpireAt:  shortLinkInput.ExpireAt,
 		CreatedBy: createdBy,
 		CreatedAt: createdAt,
 		UpdatedAt: &now,
 	}, nil
 }
 
 // NewShortLinkFake creates in memory ShortLink repository
-func NewShortLinkFake(shortLinks map[string]entity.ShortLink) ShortLinkFake {
+func NewShortLinkFake(shortLinks map[string]entity.ShortLink, userShortLinkRepoFake *UserShortLinkFake) ShortLinkFake {
```
byliuyang, line 109: Nit: prefer dependency first

backend/app/usecase/shortlink/updater.go
```diff
@@ -13,11 +11,22 @@ import (
...
 
 // ErrShortLinkNotFound represents the failure of finding certain short link in the data store
-var ErrShortLinkNotFound = errors.New("short link not found")
+type ErrShortLinkNotFound string
+
+func (e ErrShortLinkNotFound) Error() string {
+	return string(e)
+}
+
+// ErrEmptyAlias represents empty alias specified
```
byliuyang, line 20: Nit:
```suggestion
// ErrEmptyAlias represents empty alias provided.
```

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -390,7 +390,9 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
 			assert.Equal(t, testCase.expectedShortLink.LongLink, shortLink.LongLink)
 			assert.Equal(t, testCase.expectedShortLink.Alias, shortLink.Alias)
 			assert.Equal(t, testCase.expectedShortLink.CreatedAt, shortLink.CreatedAt)
-			assert.Equal(t, true, shortLink.UpdatedAt.After(now))
+			if assert.NotEqual(t, false, shortLink.UpdatedAt != nil) {
```
Coteh, line 393: I tried to do `assert.NotEqual(t, nil, shortLink.UpdatedAt)` here, but I discovered that `NotEqual` will return `true` right away if the types are different, without checking the value. See https://godoc.org/github.com/stretchr/testify/assert#NotEqual

Coteh: I am thinking about adding a wrapper to `assert.NotNil` to app, which will properly check if it's not nil. We can also update all other assertions for not nil to use `assert.NotNil` instead. What do you think?

Sorry about adding this change last second. Just wanted to make sure nil memory access can't happen here for any of the test cases if they fail.

byliuyang: @Coteh Let’s not add this one. It’s best practice to keep assertion library as simple as possible.

byliuyang: I doubt that’s why the test fails. It was working correct for me before.

backend/app/usecase/shortlink/updater_test.go
```diff
@@ -283,23 +374,28 @@ func TestShortLinkUpdaterPersist_UpdateShortLink(t *testing.T) {
...
 				assert.Equal(t, false, isExist)
 				return
 			}
 			assert.Equal(t, nil, err)
 			assert.Equal(t, testCase.expectedShortLink.LongLink, shortLink.LongLink)
 			assert.Equal(t, testCase.expectedShortLink.Alias, shortLink.Alias)
 			assert.Equal(t, testCase.expectedShortLink.CreatedAt, shortLink.CreatedAt)
-			assert.Equal(t, true, shortLink.UpdatedAt.After(now))
-			assert.Equal(t, true, userShortLinkRepo.IsRelationExist(testCase.user, shortLink))
+			if assert.NotEqual(t, false, shortLink.UpdatedAt != nil) {
```
byliuyang, line 393:
```suggestion
			if shortLink.UpdatedAt != nil {
```

PR #965, opened 2020-07-18
Add API key authenticator for third party app

byliuyang:
### Description
This enables us to expose our APIs security to third party apps and verifying their identities. In the future, we can extend this feature to check quota, perform rate limiting, and track QPS.


byliuyang:
Sorry for the large PR. Most of it is automated test.

backend/app/usecase/authenticator/thirdpartyapp.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/short-d/app/fw/crypto"
+	"github.com/short-d/app/fw/timer"
+	"github.com/short-d/short/backend/app/entity"
+	"github.com/short-d/short/backend/app/usecase/authenticator/payload"
+	"github.com/short-d/short/backend/app/usecase/authorizer"
+	"github.com/short-d/short/backend/app/usecase/keygen"
+	"github.com/short-d/short/backend/app/usecase/repository"
+)
+
+// ThirdPartyApp authenticates the identify of a third application.
```
rohithbalaji123, line 16:
```suggestion
// ThirdPartyApp authenticates the identify of a third party application.
```

backend/app/usecase/authenticator/thirdpartyapp.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	apiKeyStr := *cred.APIKey
+	tokenPayload, err := t.tokenizer.Decode(apiKeyStr)
+	if err != nil {
+		return entity.App{}, err
+	}
+
+	apiKeyPayload, err := payload.NewAPIKey(tokenPayload)
+	if err != nil {
+		return entity.App{}, errors.New("invalid api key")
```
rohithbalaji123, line 39: Just thinking if we should append the error message from `NewAPIKey` to give more context for the client.

backend/app/usecase/authenticator/thirdpartyapp.go
```diff
@@ -0,0 +1,125 @@
...
+		return "", err
+	}
+
+	_, err = t.apiKeyRepo.GetAPIKey(app.ID, key)
+	if err == nil {
+		return "", fmt.Errorf("key(%s) already exists for app(%s)", key, app.ID)
+	}
+	if !errors.As(err, &entryNotFound) {
+		return "", err
+	}
```
rohithbalaji123, line 80: I think we should add a count() method instead of making the error scenario as a control flow.

backend/app/usecase/authenticator/thirdpartyapp_test.go
```diff
@@ -0,0 +1,275 @@
...
+				"alpha": {role.Admin},
+			},
+			availableKeys:  []keygen.Key{"secret"},
+			user:           entity.User{ID: "alpha"},
+			app:            entity.App{ID: "app"},
+			expectHasErr:   true,
+			expectKeyExist: true,
+		},
+		{
+			name: "Out of available key",
```
Coteh, line 191:
```suggestion
			name: "Out of available keys",
```

backend/app/usecase/authenticator/thirdpartyapp.go
oatovar: My suggestion would be to use methods to clean up the logic using errors. Something like this could replace lines 74-80:
```go
// Example
func (t ThirdPartyApp) isAPIKeyDuplicate(appID, key string) (bool, error) {
    _, err := t.apiKeyRepo.GetAPIKey(appID, key)
    if err != nil {
        return false, err
    }
    return true, nil
}
```

backend/app/fw/must/time.go
```diff
@@ -0,0 +1,16 @@
+package must
+
+import (
+	"testing"
+	"time"
+)
+
+// Time parses time from it's string representation. This simplifies test case
+// inputs involving configurable time and normalizes it to UTC.
+func Time(t *testing.T, value string) time.Time {
```
arberiii, line 10: We should refactor or delete the function mustParseTime(https://github.com/short-d/short/blob/master/backend/app/adapter/sqldb/main_test.go#L54). Opening an issue for refactoring is sufficient for now

backend/app/fw/must/time.go
```diff
@@ -0,0 +1,16 @@
...
+	tm, err := time.Parse(time.RFC3339, value)
```
arberiii, line 11:
```suggestion
	parsedTime, err := time.Parse(time.RFC3339, value)
```

backend/app/usecase/authenticator/credential.go
```diff
@@ -0,0 +1,6 @@
+package authenticator
+
+// Credential encodes the identity of API consumer.
```
arberiii, line 3:
```suggestion
// Credential represents the identity of API consumer.
```
"Encodes" makes it look like is a function

backend/app/usecase/authorizer/rbac/permission/permission.go
```diff
@@ -4,7 +4,8 @@ package permission
 type Permission int
 
 const (
-	CreateShortLink Permission = iota
+	ViewAdminPanel Permission = iota
+	CreateShortLink
```
arberiii, line 8:
```suggestion
	ViewAdminPanel Permission = iota
	
	CreateShortLink
```

backend/app/usecase/authenticator/thirdpartyapp.go
```diff
@@ -0,0 +1,125 @@
...
+	keyGen     keygen.KeyGenerator
+	timer      timer.Timer
+	apiKeyRepo repository.APIKey
+	appRepo    repository.App
+}
+
+// GetApp retrieves app information based on the credential provided.
+func (t ThirdPartyApp) GetApp(cred Credential) (entity.App, error) {
+	if cred.APIKey == nil {
+		return entity.App{}, nil
```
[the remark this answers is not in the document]
byliuyang: Agree.

backend/app/usecase/authenticator/credential.go
byliuyang: @arberiii The JWT token is actually encoding app ID, key, and digital signature. Give this context, what are your thoughts on this?

arberiii: Then makes sense.

PR #967, opened 2020-07-19
Added missing port info in the frontend link.

MrClan:
## Current Behavior ( Optional for new feature )
Port info is missing in the generated frontend link due to which the link is not correct when the port is other than port 80.
## New Behavior
Added port info in the generated frontend link.

### DIFF
--- a/frontend/src/service/Url.service.ts
+++ b/frontend/src/service/Url.service.ts
@@ -84,7 +84,7 @@ export class UrlService {
   }
 
   aliasToFrontendLink(alias: string): string {
-    return `${window.location.protocol}//${window.location.hostname}/r/${alias}`;
+    return `${window.location.protocol}//${window.location.hostname}:${window.location.port}/r/${alias}`;
   }
 
   aliasToBackendLink(alias: string): string {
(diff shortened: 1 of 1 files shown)


rohithbalaji123:
Won't this change make the created URL look something like in the image? I am not sure if we want to show the port there. But yeah, this issue needs to be addressed.

image

MrClan:
Yes, it will show the port in the link but it should be okay since it is the correct url.

rohithbalaji123:
@MrClan , actually, `window.location.port` is empty in production site.
I guess we should use, `host` instead of `hostname`
https://stackoverflow.com/a/11379802/6533004

frontend/src/service/Url.service.ts
```diff
@@ -84,7 +84,7 @@ export class UrlService {
   }
 
   aliasToFrontendLink(alias: string): string {
-    return `${window.location.protocol}//${window.location.hostname}/r/${alias}`;
+    return `${window.location.protocol}//${window.location.hostname}:${window.location.port}/r/${alias}`;
```
rohithbalaji123, line 87:
```suggestion
    return `${window.location.protocol}//${window.location.host}/r/${alias}`;
```

MrClan:
image

The change has been verified on the testing environment and the screenshot shows that the port is not displayed in the url.

PR #973, opened 2020-07-26
Extend Table component to support resizing and aligning items

rohithbalaji123:
### Description
There is no way when using the table component to change the size of the headers or aligning the items inside the table from the parent component. These two functionalities are highly required and we should support this basic customization in a table.
### Screenshots
image

### Description
Added support to customize header font size and aligning both header and body data.
### Screenshots
image

### DIFF
--- a/frontend/src/component/ui/Table.tsx
+++ b/frontend/src/component/ui/Table.tsx
@@ -8,7 +8,7 @@ interface IProps {
   rows?: ReactChild[][];
   widths?: string[];
   alignHeaders?: TextAlignProperty[];
-  alignBody?: TextAlignProperty[];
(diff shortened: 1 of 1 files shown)


frontend/src/component/ui/Table.tsx
```diff
@@ -1,18 +1,20 @@
...
 
 import './Table.scss';
+import { TextAlignProperty } from 'csstype';
 
 interface IProps {
   headers?: ReactChild[];
   rows?: ReactChild[][];
   widths?: string[];
+  alignHeaders?: TextAlignProperty[];
+  alignBody?: TextAlignProperty[];
```
byliuyang, line 11:
```suggestion
  alignBodyColums?: TextAlignProperty[];
```

Issue #977, opened 2020-07-27
[Refactor] Replace mustParseTime with must.Time across the codebase


byliuyang:
**What is frustrating you?**
The existing tests use [mustParseTime](https://github.com/short-d/short/blob/master/backend/app/adapter/sqldb/main_test.go#L54) helper to convert human readable time in the test case to help test readers understand the test case. However, it does NOT inherently support returning `Time pointer`, preventing time string being directly embed into the test case. https://github.com/short-d/short/pull/965 is adding `must.Time()` to support point return value. This lead to code duplication & inconsistency.

**Your solution**
Remove or refactor `mustParseTime` across the codebase.

PR #980, opened 2020-07-29
Replace mustParseTime with must.Time across the codebase

arberiii:
Fixes [issue #977](https://github.com/short-d/short/issues/977)

### DIFF
--- a/backend/app/adapter/sqldb/app_integration_test.go
+++ b/backend/app/adapter/sqldb/app_integration_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/short-d/short/backend/app/adapter/sqldb"
 	"github.com/short-d/short/backend/app/adapter/sqldb/table"
 	"github.com/short-d/short/backend/app/entity"
+	"github.com/short-d/short/backend/app/fw/must"
 )
 
 var insertAppRowSQL = fmt.Sprintf(`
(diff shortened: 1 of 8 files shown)


backend/app/adapter/sqldb/app_integration_test.go
```diff
@@ -31,7 +32,7 @@ type appTableRow struct {
 }
 
 func TestAppSQL_FindAppByID(t *testing.T) {
-	createdAt := mustParseTime(t, "2017-05-01T08:02:16-07:00")
+	createdAt := must.Time(t, "2017-05-01T08:02:16-07:00")
```
byliuyang, line 35: We need to place all the `createdAt` variable in the test case themselves.

backend/app/adapter/sqldb/short_link_integration_test.go
```diff
@@ -277,8 +278,8 @@ func TestShortLinkSql_IsAliasExist(t *testing.T) {
 }
 
 func TestShortLinkSql_GetShortLinkByAlias(t *testing.T) {
-	twoYearsAgo := mustParseTime(t, "2017-05-01T08:02:16-07:00")
-	now := mustParseTime(t, "2019-05-01T08:02:16-07:00")
+	twoYearsAgo := must.Time(t, "2017-05-01T08:02:16-07:00")
```
byliuyang, line 281: We should create a helper in ptr package which converts `time.Time` to `*time.Time` and embed the parsing directly in the test cases to improve readability.

byliuyang: We should do this for all test cases referenced in this PR!

Issue #599, opened 2020-03-23
[REFACTOR] Simplify invokeCreateShortLinkApi() to improve readability


JackieYJC:
Currently the code: https://github.com/short-d/short/blob/master/frontend/src/service/Url.service.ts#L132-L166 is difficult to read and maintain.

Try to simplify the business logic to improve the readability, in terms of variable names, method abstractions, and type clarity.

PR #985, opened 2020-07-31
Refactor invokeCreateShortLinkApi 

Dean430:
This is for issue https://github.com/short-d/short/issues/599
Refactor invokeCreateShortLinkApi.


Dean430:
I noticed @byliuyang codecov/patch — 0.00% of diff hit (target 52.28%) is failing. Looks like we don't have any unit test for the ts files?

byliuyang:
@Dean430 You are correct. Let's add unit test for next PR.

PR #992, opened 2020-08-04
Use ptr.String for test cases that contain strings

### LINKED ISSUE #959: [Refactor] Use ptr.String to pass test case strings directly into test cases
**What is frustrating you?**
Test case strings should be inserted directly into test cases if possible, but [due to a restriction in the Golang spec](https://golang.org/ref/spec#Address_operators), string literals cannot have its address taken directly.

**Your solution**
In #938, `ptr.String` method was introduced to facilitate getting address out of string literal in a clean way. Update all test cases that use string addresses to use this helper method instead.

Coteh:
Fixes #959. Use `ptr.String()` to allow for direct embedding of strings into tests.


byliuyang:
Drone CI was broken. Merging now.

PR #994, opened 2020-08-15
Add localhost for swagger docs

Ashakibp:
## Current Behavior ( Optional for new feature )
We currently don't have a localhost option on the swagger doc for running API calls

## New Behavior

This PR will add a localhost option to the doc template.


backend/app/adapter/routing/api.yml
```diff
@@ -15,6 +15,7 @@ servers:
   - url: https://api.short-d.com
   - url: https://api-staging.short-d.com
   - url: https://api-testing.short-d.com
+  - url: https://localhost.com
```
byliuyang, line 18:
```suggestion
  - url: http://localhost
```
We don't use HTTPS for local servers because we are not giving developers our private certificate needed for TLS.

Ashakibp: gotchya

PR #996, opened 2020-08-16
Upgrade to Storybook 6

Coteh:
There's some breaking changes and deprecations that I'm still sorting out so leaving this in draft for now.

Using this guide: https://medium.com/storybookjs/storybook-6-migration-guide-200346241bb5

Checklist:
- [X] Remove `withKnobs` decorator since it's now baked in with its preset ([source](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#new-addon-presets))
- [x] Remove TypeScript configuration since there's now a baked in default ([source](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#zero-config-typescript))
- [x] Switch `addons-info`, which is now deprecated, to  `addons-docs` ([source](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#60-deprecations))
- [ ] Consider switching `addons-knobs` to [Storybook Controls](https://medium.com/storybookjs/storybook-controls-ce82af93e430)
- [ ] Convert story items to use Storybook Args ([source](https://medium.com/storybookjs/introducing-storybook-args-2dadcdb777cc))
- [ ] Ensure everything else from Migrations guide is checked off
- [ ] Consider adding MDX documentation to stories (should probably make this a separate issue that can be dealt with later, since it's more of a nice-to-have) ([source](https://github.com/storybookjs/storybook/tree/next/addons/docs#mdx))
- [x] Bump to 6.0.12
- [ ] Bump to newest minor version before pushing


story/.storybook/main.js
```diff
@@ -1,26 +1,25 @@
...
+        {
+          loader: 'css-loader',
+          options: {
+            modules: true
+          }
+        },
+        'sass-loader'
+      ]
+    });
+    config.resolve.extensions.push('.ts', '.tsx', '.scss');
```
byliuyang, line 22: Are `.ts` and `.tsx` still needed here?

story/src/0-Button.stories.tsx
```diff
@@ -1,15 +1,13 @@
...
-import { text, withKnobs } from '@storybook/addon-knobs';
-import { withInfo } from '@storybook/addon-info';
+import { text } from '@storybook/addon-knobs';
 
 import styles from './0-Button.stories.module.scss';
 
 export default {
   title: 'UI/Button',
-  component: <Button />,
-  decorators: [withKnobs, withInfo({ header: false, inline: true })]
```
byliuyang, line 12: Is the doc style still looking correct?

Coteh: Yes, my understanding is that with Storybook Docs, even though it's built to support the new MDX format, you can still use Component Story Format (CSF).

I did notice that it seems as though you have to pass in the component type rather than a JSX element into the `component` field now in order to get the knobs to work. This is how they seem to do it [here](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/docspage.md) too.

story/.storybook/main.js
Coteh: Good question, I'll check this. 👍

PR #1000, opened 2020-08-21
Emit errors through Search API

### LINKED ISSUE #865: [Feature] Handle errors of Search API
**What is the problem?**
In the first version of [Search API ](https://github.com/short-d/short/tree/master/backend/app/usecase/search)we are not handling errors and for now, we are just logging them.

**Your solution**
Handle errors correctly and ideally use [instrumentation code](https://github.com/short-d/short/blob/master/backend/app/usecase/instrumentation/instrumentation.go) so that we have a centralized place to organize everything and make changes.

Coteh:
Fixes #865 

Tests not added yet, so this will be in draft until they're added.

Emit errors through search API and add new types for unknown resource and user not provided.

TODO
- [x] Add test for user not provided
- [x] Add tests for unknown resource
  - [x] one unknown resource
  - [x] ~two unknown resources~ edit: redundant, other two cover it
  - [x] one unknown resource and one known resource
- [x] Decide whether errors should be emitted as text or as JSON
- [x] Decide whether to fix two warnings reported by GoLand in `handle/search.go` (edit: going to skip them, beyond the scope of this PR imo)

### DIFF
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -168,7 +168,7 @@ func (f *Filter) resourcesString() []string {
 	return resources
 }
 
-func newSearchResponse(result search.Result) SearchResponse {
+func newSearchResponse(result search.ResourceResult) SearchResponse {
 	shortLinks := make([]ShortLink, len(result.ShortLinks))
 	for i := 0; i < len(result.ShortLinks); i++ {
 		shortLinks[i] = newShortLink(result.ShortLinks[i])
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -174,7 +174,7 @@ func (f *Filter) resourcesString() []string {
 	return resources
 }
 
-func newSearchResponse(result search.ResourceResult) SearchResponse {
+func newSearchResponse(result search.Result) SearchResponse {
 	shortLinks := make([]ShortLink, len(result.ShortLinks))
 	for i := 0; i < len(result.ShortLinks); i++ {
 		shortLinks[i] = newShortLink(result.ShortLinks[i])
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -213,8 +213,8 @@ func newUser(user entity.User) User {
 }
 
 func emitSearchError(w http.ResponseWriter, err error) {
-	var code = http.StatusInternalServerError
 	var (
+	       code http.StatusInternalServerError
 		u search.ErrUserNotProvided
 		r search.ErrUnknownResource
 	)
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -214,9 +214,9 @@ func newUser(user entity.User) User {
 
 func emitSearchError(w http.ResponseWriter, err error) {
 	var (
-	       code http.StatusInternalServerError
-		u search.ErrUserNotProvided
-		r search.ErrUnknownResource
+		code = http.StatusInternalServerError
+		u    search.ErrUserNotProvided
+		r    search.ErrUnknownResource
 	)
 	if errors.As(err, &u) {
 		code = http.StatusUnauthorized
--- a/backend/app/adapter/routing/handle/search.go
+++ b/backend/app/adapter/routing/handle/search.go
@@ -2,6 +2,7 @@ package handle
 
 import (
 	"encoding/json"
+	"errors"
 	"io/ioutil"
 	"net/http"
 	"time"
@@ -47,6 +48,11 @@ type SearchResponse struct {
 	Users      []User      `json:"users,omitempty"`
 }
 
+// SearchError represents an error with the Search API request.
+type SearchError struct {
+	Message string `json:"message"`
+}
+
 // ShortLink represents the short_link field of Search API respond.
 type ShortLink struct {
 	Alias     string     `json:"alias,omitempty"`
@@ -79,7 +85,7 @@ func Search(
 		defer r.Body.Close()
 		if err != nil {
 			i.SearchFailed(err)
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			emitSearchError(w, err)
 			return
 		}
 
@@ -87,7 +93,7 @@ func Search(
 		err = json.Unmarshal(buf, &body)
 		if err != nil {
 			i.SearchFailed(err)
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			emitSearchError(w, err)
 			return
 		}
 
@@ -99,14 +105,14 @@ func Search(
 		filter, err := search.NewFilter(body.Filter.MaxResults, body.Filter.Resources, body.Filter.Orders)
 		if err != nil {
 			i.SearchFailed(err)
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			emitSearchError(w, err)
 			return
 		}
 
 		results, err := searcher.Search(query, filter)
 		if err != nil {
 			i.SearchFailed(err)
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			emitSearchError(w, err)
 			return
 		}
 
@@ -114,7 +120,7 @@ func Search(
 		respBody, err := json.Marshal(&response)
 		if err != nil {
 			i.SearchFailed(err)
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			emitSearchError(w, err)
 			return
 		}
 
@@ -204,4 +210,25 @@ func newUser(user entity.User) User {
 		CreatedAt:      user.CreatedAt,
 		UpdatedAt:      user.UpdatedAt,
 	}
+}
+
+func emitSearchError(w http.ResponseWriter, err error) {
+	var code = http.StatusInternalServerError
+	var (
+		u search.ErrUserNotProvided
+		r search.ErrUnknownResource
+	)
+	if errors.As(err, &u) {
+		code = http.StatusUnauthorized
+	}
+	if errors.As(err, &r) {
+		code = http.StatusNotFound
+	}
+	errResp, err := json.Marshal(SearchError{
+		Message: err.Error(),
+	})
+	if err != nil {
+		return
+	}
+	http.Error(w, string(errResp), code)
 }
--- a/backend/app/usecase/search/search.go
+++ b/backend/app/usecase/search/search.go
@@ -21,12 +21,6 @@ type Search struct {
 
 // Result represents the result of a search query.
 type Result struct {
-	Resources ResourceResult
-	Err       error
-}
-
-// ResourceResult represents the resources obtained from a search query.
-type ResourceResult struct {
 	ShortLinks []entity.ShortLink
 	Users      []entity.User
 }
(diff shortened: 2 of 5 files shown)


backend/app/adapter/routing/handle/search.go
```diff
@@ -79,15 +85,15 @@ func Search(
 		defer r.Body.Close()
```
Coteh, line 85: GoLand complains about error not being handled here. I am thinking this can be handled in a separate PR, what do you guys think?

backend/app/adapter/routing/handle/search.go
```diff
@@ -99,22 +105,22 @@ func Search(
...
 		}
 
 		response := newSearchResponse(results)
 		respBody, err := json.Marshal(&response)
 		if err != nil {
 			i.SearchFailed(err)
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			emitSearchError(w, err)
 			return
 		}
```
Coteh, line 126: GoLand complains about error not being handled here (the `w.Write` call on line 127). I am thinking this can be handled in a separate PR, what do you guys think?

arberiii: This error can be only network error or some internal error, so I think the best option is to just log it and not handle

backend/app/adapter/routing/handle/search.go
```diff
@@ -205,3 +211,24 @@ func newUser(user entity.User) User {
 		UpdatedAt:      user.UpdatedAt,
 	}
 }
+
+func emitSearchError(w http.ResponseWriter, err error) {
+	var code = http.StatusInternalServerError
+	var (
+		u search.ErrUserNotProvided
+		r search.ErrUnknownResource
+	)
```
arberiii:
```suggestion
	var (
	       code http.StatusInternalServerError
		u search.ErrUserNotProvided
		r search.ErrUnknownResource
	)
```

backend/app/usecase/search/search.go
```diff
@@ -22,12 +21,32 @@ type Search struct {
...
+	Err       error
+}
+
+// ResourceResult represents the resources obtained from a search query.
+type ResourceResult struct {
 	ShortLinks []entity.ShortLink
 	Users      []entity.User
 }
 
+// ErrUnknownResource represents unknown search resource error.
```
arberiii, line 28: Maybe we can move this new error.go file. Wdty?

backend/app/usecase/search/search.go
```diff
@@ -22,12 +21,32 @@ type Search struct {
...
+func (e ErrUnknownResource) Error() string {
+	return string(e)
+}
+
+// ErrUserNotProvided represents user not provided for search query.
+type ErrUserNotProvided string
+
+func (e ErrUserNotProvided) Error() string {
+	return string(e)
+}
```
arberiii:
```suggestion
type ErrUserNotProvided struct{}

func (e ErrUserNotProvided) Error() string {
	return "user not provided"
}
```
What do you think?

arberiii (approved):
@Coteh Thank you so much for working in emitting errors!

backend/app/adapter/routing/handle/search.go
Coteh: Should the handler still exit at this point though? Also, what would the HTTP response look like in this case? I believe that if I remove the handler code it will just return an empty response. Is this OK?

backend/app/usecase/search/search.go
Coteh: In other usecase files, such as [creator.go](https://github.com/short-d/short/blob/00df27fbd6d2abda4e2c54a76110577d1fff413d/backend/app/usecase/shortlink/creator.go) the error types are embedded within the file. I am not sure what the benefit of having it in a separate file would be, especially since there's only two custom error types at the moment.

backend/app/usecase/search/search.go
```diff
@@ -38,50 +57,63 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
...
+func (s Search) searchResource(resource Resource, orderBy order.Order, query Query, filter Filter) (ResourceResult, error) {
 	switch resource {
 	case ShortLink:
 		return s.searchShortLink(query, orderBy, filter)
 	case User:
 		return s.searchUser(query, orderBy, filter)
 	default:
-		return Result{}, errors.New("unknown resource")
+		// TODO add tests for unknown resource
+		return ResourceResult{}, ErrUnknownResource("unknown resource")
```
Coteh:
```suggestion
		return ResourceResult{}, ErrUnknownResource{}
```

backend/app/usecase/search/search.go
```diff
@@ -38,50 +57,63 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
...
 }
 
 // TODO(issue#866): Simplify searchShortLink function
-func (s Search) searchShortLink(query Query, orderBy order.Order, filter Filter) (Result, error) {
+func (s Search) searchShortLink(query Query, orderBy order.Order, filter Filter) (ResourceResult, error) {
 	if query.User == nil {
-		s.logger.Error(errors.New("user not provided"))
-		return Result{}, nil
+		// TODO add a test for user not provided
+		err := ErrUserNotProvided("user not provided")
```
Coteh:
```suggestion
		err := ErrUserNotProvided{}
```

backend/app/usecase/search/search.go
Coteh: LGTM, I'll update the usage as well as the unknown resource error type to match.

backend/app/usecase/search/search.go
```diff
@@ -22,12 +21,32 @@ type Search struct {
...
 	ShortLinks []entity.ShortLink
 	Users      []entity.User
 }
 
+// ErrUnknownResource represents unknown search resource error.
+type ErrUnknownResource string
+
+func (e ErrUnknownResource) Error() string {
+	return string(e)
+}
```
Coteh:
```suggestion
type ErrUnknownResource struct{}

func (e ErrUnknownResource) Error() string {
	return "unknown resource"
}
```

backend/app/adapter/routing/handle/search.go
naturetime10: This is fine. Please ignore this one.

backend/app/adapter/routing/handle/search.go
naturetime10: Please ignore write error for now.

byliuyang:
Please hold this PR. We need to discuss what are the expected behaviors.

backend/app/adapter/routing/handle/search.go
```diff
@@ -47,6 +48,11 @@ type SearchResponse struct {
 	Users      []User      `json:"users,omitempty"`
 }
 
+// SearchError represents an error with the Search API request.
+type SearchError struct {
+	Message string `json:"message"`
+}
```
naturetime10, line 54: I wonder what is this for? Are we going handle it in the frontend?

backend/app/adapter/routing/handle/search.go
```diff
@@ -205,3 +211,24 @@ func newUser(user entity.User) User {
...
+		code = http.StatusNotFound
+	}
+	errResp, err := json.Marshal(SearchError{
+		Message: err.Error(),
+	})
+	if err != nil {
+		return
+	}
+	http.Error(w, string(errResp), code)
+}
```
naturetime10, line 234: Each search error is different in the code. I recommend putting the content of this function back to the code so that the reader can follow the error handling logic. This is an example when abstraction is reducing readability.

backend/app/usecase/search/search.go
```diff
@@ -22,12 +21,32 @@ type Search struct {
...
+// ErrUserNotProvided represents user not provided for search query.
+type ErrUserNotProvided struct{}
+
+func (e ErrUserNotProvided) Error() string {
+	return "user not provided"
+}
+
 // Search finds resources based on specified criteria.
-func (s Search) Search(query Query, filter Filter) (Result, error) {
+func (s Search) Search(query Query, filter Filter) (ResourceResult, error) {
```
naturetime10: We are already returning error here. I wonder why are we putting extract error in `ResourceResult`. Go is making error an interface. By design, the language wants us to return whatever error we get directly.

backend/app/usecase/search/search.go
```diff
@@ -38,50 +57,61 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
 		go func() {
 			result, err := s.searchResource(filter.resources[i], orders[i], query, filter)
 			if err != nil {
-				// TODO(issue#865): Handle errors of Search API
 				s.logger.Error(err)
-				resultCh <- Result{}
+				resultCh <- Result{
+					Resources: ResourceResult{},
+					Err:       err,
+				}
```
naturetime10:
```suggestion
We can send error into an error channel and handle it inside the `select` statement below.
```

backend/app/usecase/search/search.go
```diff
@@ -38,50 +57,61 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
...
+				Resources: result,
+				Err:       nil,
+			}
 		}()
 	}
 
 	timeout := time.After(s.timeout)
-	var results []Result
+	var results []ResourceResult
+	var resultErr error
```
naturetime10, line 68: This is not needed based on the previous comment.

backend/app/usecase/search/search.go
```diff
@@ -38,50 +57,61 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
...
-func (s Search) searchResource(resource Resource, orderBy order.Order, query Query, filter Filter) (Result, error) {
+func (s Search) searchResource(resource Resource, orderBy order.Order, query Query, filter Filter) (ResourceResult, error) {
 	switch resource {
 	case ShortLink:
 		return s.searchShortLink(query, orderBy, filter)
 	case User:
 		return s.searchUser(query, orderBy, filter)
 	default:
-		return Result{}, errors.New("unknown resource")
+		return ResourceResult{}, ErrUnknownResource{}
```
naturetime10: Let's revert this. Didn't seems to provide value here.

backend/app/usecase/search/search.go
```diff
@@ -38,50 +57,61 @@ func (s Search) Search(query Query, filter Filter) (Result, error) {
...
 }
 
 // TODO(issue#866): Simplify searchShortLink function
-func (s Search) searchShortLink(query Query, orderBy order.Order, filter Filter) (Result, error) {
+func (s Search) searchShortLink(query Query, orderBy order.Order, filter Filter) (ResourceResult, error) {
 	if query.User == nil {
-		s.logger.Error(errors.New("user not provided"))
-		return Result{}, nil
+		err := ErrUserNotProvided{}
+		s.logger.Error(err)
```
naturetime10: Let's directly return error here and log it at the caller.

naturetime10 (changes requested):

backend/app/adapter/routing/handle/search.go
Coteh: This will provide information about the search error that occurred. So far, we have "user not provided" and "unknown resource" errors. I think it would make sense to emit these errors in the frontend, even though they may only show up in rare circumstances (e.g. user somehow gets logged out right before they search for something). What do you think?

backend/app/usecase/search/search.go
Coteh: I think this is still needed, because we would like to return only the first error that occurred. What do you think?

Issue #954, opened 2020-07-12
[Feature] Add UI components to storybook


Coteh:
Add storybook items for the various UI components we have within short.

| Component        | Status |
| ------------- | ------------- |
| Button | Completed (https://github.com/short-d/short/pull/943) |
| ChangeLogModal | Not Started |
| Drawer | Not Started |
| Icon | Not Started |
| MarkdownViewer | Not Started |
| Modal | Not Started |
| Navigation | Not Started |
| Notice | Not Started |
| PageControl | Not Started |
| SearchBar | Not Started |
| Section | Not Started |
| Spinner | Not Started |
| Table | Not Started |
| Tabs | Not Started |
| Toast | Not Started |
| Toggle | Not Started |
| ViewChangeLogButton | Not Started |
| Auth Loading Page Spinner (https://github.com/short-d/short/issues/816) | Not Started |

PR #1001, opened 2020-08-22
Add story for Table component

Coteh:
Addresses the `Table` story in #954.

**Leaving in Draft until these items are done:**
- [x] Complete knobs for columns widths and header font size
- [x] Determine whether it's necessary to have knobs for align headers and align body columns
- [ ] Decide whether to transition from knobs to Storybook Controls **edit: will address in separate PR**
  - Learn more [here](https://medium.com/storybookjs/storybook-controls-ce82af93e430), I think we should move over to Controls because it seems to be the best practice for Storybook 6 and going forward.
  - [I am currently working on adding Storybook Controls in this branch](https://github.com/short-d/short/tree/storybook-controls) (not ready for PR yet)

This PR also adds Storybook Backgrounds, as I wanted to add the color #fafafa, which is being used underneath the Favourite short links table in the homepage, as a background colour for the canvas. This will make the `Table` component stand out better. I also set #fafafa as the default background colour for story canvases.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 [![codecov](https://codecov.io/gh/short-d/short/branch/master/graph/badge.svg)](https://codecov.io/gh/short-d/short)
 [![Maintainability](https://api.codeclimate.com/v1/badges/910f974653f1b3495534/maintainability)](https://codeclimate.com/github/short-d/short/maintainability)
 [![Go Report Card](https://goreportcard.com/badge/github.com/short-d/short)](https://goreportcard.com/report/github.com/short-d/short)
-[![Open Source Love](https://badges.frapsoft.com/os/mit/mit.svg?v=102)](https://github.com/byliuyang/short)
+[![MIT License](https://img.shields.io/badge/License-MIT-blue.svg)](https://github.com/byliuyang/short)
 [![Floobits Status](https://floobits.com/byliuyang/short.svg)](https://floobits.com/byliuyang/short/redirect)
 
 ![Demo](promo/marquee.png)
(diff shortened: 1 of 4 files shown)


byliuyang:
@Coteh Control seems good. Let's do it!

Coteh:
@byliuyang I think what I'll do is merge this in as-is, then make a separate PR for Controls and convert all existing stories to use Controls in that PR. What do you think?

story/src/2-Table.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export const table = () => {
+  const headers = array('Header Columns', ['Long Link', 'Alias']);
+  return (
+    <Table
+      headers={headers}
+      rows={new Array(number('Number of rows', 2))
+        .fill(0)
+        .map((_, i) => headers.map((_, j) => `data ${i}-${j}`))}
```
byliuyang, line 33: I am confused by what's going on here. Can you explain what's going on or extract the logic using helper functions?

story/.storybook/preview.js
```diff
@@ -0,0 +1,15 @@
...
+        name: 'white',
+        value: '#fff'
+      },
+      {
+        name: 'short',
+        value: '#fafafa'
+      }
+    ]
+  }
+};
```
byliuyang, line 15: I wonder what are those?

story/src/2-Table.stories.tsx
Coteh: I am simply allowing the table to have the number of rows user specifies within the storybook. I can make it into a helper function if that makes it easier to understand.

story/.storybook/preview.js
Coteh: These are the parameters for storybook backgrounds https://storybook.js.org/docs/react/essentials/backgrounds

Coteh:
@byliuyang Screenshot added

PR #1006, opened 2020-08-27
Add Spinner story

Coteh:
image


frontend/src/component/ui/Spinner.tsx
```diff
@@ -1,19 +1,19 @@
...
-      <div className={'spinner'}>
-        <div className={'solar-system'}>
-          <div className={'earth-orbit orbit'}>
-            <div className={'planet earth'} />
-            <div className={'venus-orbit orbit'}>
-              <div className={'planet venus'} />
-              <div className={'mercury-orbit orbit'}>
-                <div className={'planet mercury'} />
-                <div className={'sun'} />
+      <div className={styles['spinner']}>
```
naturetime10:
```suggestion
      <div className={styles.spinner}>
```

frontend/src/component/ui/Spinner.tsx
```diff
@@ -1,19 +1,19 @@
...
+        <div className={styles['solar-system']}>
```
naturetime10, line 9:
```suggestion
        <div className={styles.solarSystem}>
```
Does this work?

Coteh: No it does not. I would have to rename the class to `solarSystem` in the SCSS module.

naturetime10: @Coteh Okey. I will let the PR go through this time. We will have to figure out a styling guide in the future.

story/src/2-Spinner.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Spinner } from '../../frontend/src/component/ui/Spinner';
+
+export default {
+  title: 'UI/Spinner',
+  component: Spinner
+};
+
+export const spinner = () => {
+  return <Spinner />;
```
naturetime10: Can we wrap the spanner with a background? It's too to see the orbitals and planet on storybook.

byliuyang (dismissed):

story/src/2-Spinner.stories.tsx
Coteh: I'll see what I can do with Storybook Backgrounds. I'll wrap the component with background as fallback in case that doesn't work out.

Coteh: Here's how it looks now, using the dark color that comes with Storybook Backgrounds addon

image

Coteh: One caveat, if you switch background color, it'll always stay that color, even when selecting `Spinner` story. I'm going to fallback to wrapping.

Coteh: image

